PR #3, opened 2021-04-21
Ccho terms recommendations

ccho-mongodb:
DO NOT MERGE

Staging:
https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/ccho-terms-recommendations/terms/

### DIFF
--- a/source/terms.txt
+++ b/source/terms.txt
@@ -13,7 +13,9 @@ no other reasonable interpretation. Consult the documentation in the
 
 .. list-table::
    :header-rows: 1
-
+   :stub-columns: 1 
+   :widths: 24 38 38
+   
    * - Term
      - Description
      - Recommendation
@@ -68,8 +70,6 @@ no other reasonable interpretation. Consult the documentation in the
      - data
      - See https://golang.org/ref/spec#Assignability
        https://golang.org/ref/spec#Representability
-
-     - See https://golang.org/ref/spec#Type_identity
       
    * - **argument**
      - data you pass to a function or method that corresponds to one
@@ -114,7 +114,7 @@ no other reasonable interpretation. Consult the documentation in the
 
    
 
-.. _golang-terms-sources
+.. _golang-terms-sources:
 
 Sources
 -------
--- a/source/terms.txt
+++ b/source/terms.txt
@@ -1,0 +1,126 @@
+==============================
+Go Terminology Recommendations
+==============================
+
+.. default-domain:: mongodb
+
+
+This is a subset of terms and situations in which we may need to use
+throughout our Go Driver documentation. We should use these when
+appropriate to disambiguate terms, but do not need to use them if there is
+no other reasonable interpretation. Consult the documentation in the
+:ref:`Sources <golang-terms-sources>` for additional guidance.
+
+.. list-table::
+   :header-rows: 1
+
+   * - Term
+     - Description
+     - Recommendation
+
+   * - **package**
+     - a list of struct, function, and/or method in a file, identified by
+       the "package <name>" at the top of the file.
+     -
+
+   * - **module**
+     - a collection of packages in the same directory, described by
+       a go.mod file at the directory root.
+     -
+   * - **Function**
+     - Declared in a package using ``func`` and does not specify a receiver.
+     - Refer to this as a "function", omitting parameters. Feel free to
+       use appropriate modifiers for specificity such as "variadic". The
+       function in the example below is ``NewColor()``:
+
+       .. code-block:: go
+
+          func NewColor(name string) *Color {}
+
+       Common usage:
+
+       - "Call the ``NewColor()`` function"
+
+   * - **Method**
+     - A function declared with a receiver.
+     - Refer to this as a "method", omitting parameters, and ensure you
+       include context on what the receiver is. Feel free to use
+       appropriate modifiers for specificity such as "variadic". The method
+       in the example below is ``isPrimary()`` and the receiver is ``Color``.
+
+       .. code-block:: go
+
+          func (a *Color) isPrimary() bool {}
+
+       This is how the method might be called:
+
+       .. code-block:: go
+
+          // assume assignment c := NewColor("red")
+          c.isPrimary()
+
+   * - **(return/result) values**
+     - The data returned from a method or function. You can find 
+       signature.
+     -
+
+   * - **value**
+     - data
+     - See https://golang.org/ref/spec#Assignability
+       https://golang.org/ref/spec#Representability
+
+     - See https://golang.org/ref/spec#Type_identity
+      
+   * - **argument**
+     - data you pass to a function or method that corresponds to one
+       of its parameters.
+     - Use this term when you need to describe the data you're passing without
+       specific details like type, usually to refer to position in the
+       method/function signature.
+
+       Here are a few examples of how to use this term:
+       - "Pass your connection string as the second argument to
+         ``mongo.Connect()``"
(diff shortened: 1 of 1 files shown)


source/terms.txt
```diff
@@ -0,0 +1,169 @@
...
+          func (a *Color) isPrimary() bool {}
+
+       | "You can call the ``isPrimary()`` method on a ``Color`` pointer receiver type to determine whether the color is red, yellow, or blue as shown below:"
+
+       .. code-block:: go
+
+          // assume assignment c := NewColor("red")
+          c.isPrimary()
+
+   * - **return values/lists**
```
biniona-mongodb, line 69: Are the term "return values" and "return lists" considered equivalent? 

If I were to use this comment in writing I take it to mean "return value" is the singular form, and "return list" is the plural form. Is that accurate?

source/terms.txt
```diff
@@ -0,0 +1,169 @@
...
+       | Here are a few examples of how to use this term:
+       - "Pass your connection string as the second argument to ``mongo.Connect()``"
+       - "The ``NewColor()`` function takes a ``string`` argument."
+       - `Golang.org reference docs <https://golang.org/ref/spec#Passing_arguments_to_..._parameters>`__
+
+   * - **value** (passed)
+     - | **Description:**
+       | Similar to the `wikipedia definition of value <https://en.wikipedia.org/wiki/Value_(computer_science)>`__ in that it is the "value of a variable" that can be of any data type. This is different from **return values/lists** in that it refers to the contents of the type. See `Go by Example: Values <https://gobyexample.com/values>`__ for examples.
+       | **Recommendation:**
+       | **TODO:** decide whether this term should be formatted in monospace, quotes, or both.
```
biniona-mongodb, line 90: Good question. Are there any other terms (not in-code examples of terms) that receive special formatting? 

I can't think of any of the top of my head, so my gut is that the term value should not receive special formatting. I am open to an argument that it should however.

source/terms.txt
```diff
@@ -0,0 +1,169 @@
...
+
+       | When referring to result parameters (also called identifiers), use **result** or **result list**. Example(s):
+
+       - "``func ChangeColor(newColor string) (string, error)``. This function returns a result list containing a ``string`` and ``error``."
+
+   * - **type**
+     - | **Description:**
+       | "A type determines a set of values together with operations and methods specific to those values" (`Types <https://golang.org/ref/spec#Types>`__)
+       | **Recommendation:**
+       | Use this when describing a data type, including categorical (pointer, function, strcut, array, string, etc.). The term "type" should not receive any special formatting, but a specific type name should be monospaced.
```
biniona-mongodb, line 110:
```suggestion
       | Use this when describing a data type, including categorical (pointer, function, strcut, array, string, etc.). The term "type" should not receive any special formatting, but a specific type name should be monospaced. For example, "the ``int`` type."
```

I thought this point has a natural example so I wanted to include it.

source/terms.txt
```diff
@@ -0,0 +1,169 @@
...
+
+   * - **pointer**
+     - | **Description:**
+       | References to values/records. See `Go by Example: Pointers <https://gobyexample.com/pointers>`__ for further explanation and examples.
+       | **Recommendation:**
+       | We can talk about pointers as ``*C`` or "pointer to ``C``". The latter may be easier for a new developer to understand, and to discover the meaning through a google search.
```
biniona-mongodb, line 116: Would it be possible to take a firm stance on which of these we will use going forward? I would be fine only using "pointer to ``C``" for all references to pointers in prose.

source/terms.txt
```diff
@@ -0,0 +1,169 @@
...
+            name string
+            r,g,b int
+          }
+
+       | **Recommendation:**
+       | Use struct or struct type to refer to both the declaration within a package as well as the data held by a variable.
+       | Example(s):
+
+       - "The color struct type contains three ``int`` fields."
+       - ``c := color(name: "crimson", r: 153, g: 0, b: 0)`` Call ``isPrimary() on the ``color`` struct to determine whether it is a primary color.
```
biniona-mongodb, line 135:
```suggestion
       - ``c := color(name: "crimson", r: 153, g: 0, b: 0)`` Call ``isPrimary()`` on the ``color`` struct to determine whether it is a primary color.
```

source/terms.txt
```diff
@@ -0,0 +1,169 @@
...
+
+       .. code-block:: go
+
+          type color struct {
+            name string
+            r,g,b int
+          }
+
+       | **Recommendation:**
+       | Use struct or struct type to refer to both the declaration within a package as well as the data held by a variable.
```
biniona-mongodb, line 131: Would we be alright drawing a hard line for when one is used over the other?

I would be fine using only "struct type" to  refer to the data held in a variable. And only using "struct" to refer to a declaration within a package.

source/terms.txt
```diff
@@ -0,0 +1,169 @@
...
+       | Example(s):
+
+       - "The color struct type contains three ``int`` fields."
+       - ``c := color(name: "crimson", r: 153, g: 0, b: 0)`` Call ``isPrimary() on the ``color`` struct to determine whether it is a primary color.
+
+   * - **variable**
+     - | **Description:**
+       | Holds data whether a data structure or pointer type and is accessed by the name by which it was declared. See `Variables <https://golang.org/ref/spec#Variables>`__ for more info. The following shows a few example variable declarations:
```
biniona-mongodb, line 139:
```suggestion
       | Holds data, whether a data structure or pointer type, and is accessed by the name by which it was declared. See `Variables <https://golang.org/ref/spec#Variables>`__ for more info. The following shows a few example variable declarations:
```
I had trouble understanding this sentence without commas

source/terms.txt
```diff
@@ -0,0 +1,169 @@
...
+
+       .. code-block:: go
+
+          var c = color.NewColor("chartreuse")
+          x := color.NewColor("indigo")
+          _, g, _ := c.RGBValues()
+          var pColor *Color
+
+       | **Recommendation:**
+       | Use "variable" when you are referencing one but do not require any specificity. If it is more helpful to refer to the variable by its name, use that instead.
```
biniona-mongodb, line 149:
```suggestion
       | Use "variable" when you are referencing one but do not require any specificity. If it is more helpful to refer to the variable by its name, use that instead when possible.
```

Is there a situation where you don't have a variable to refer to?

source/terms.txt
```diff
@@ -0,0 +1,169 @@
...
+       | Example:
+
+       - "You can store the return value in a value so you can access it later."
+
+   * - **pointer reciever**
+     - | **Description:**
+       | "You can declare methods with pointer receivers. This means the receiver type has the literal syntax ``*T`` for some type ``T``. (Also, ``T`` cannot itself be a pointer such as ``*int``.)" (from `Pointer receivers <https://tour.golang.org/methods/4>`__.
+       | **Recommendation:**
+       | We decided as a team that no further specificity is needed when using this term in prose, but can use the ``*`` prefix when it is featured in a list of links to API documentation.
+
```
biniona-mongodb, line 159: I was surprised a definition for object was not included. Are we avoiding the term "object" in our go documentation? I support not using "object", but I thought you were in favor of using object (as it is technically correct). 

I am only writing this comment, as this was surprising given how I understood your stance on the term "object" in golang.

biniona-mongodb (approved):
Excellent job! This is amazing. I would 100% support putting this in a `meta` tab in the main TOC. I think this could be really helpful for readers and writers, and would give us a natural place to put definitions of terms as we define them throughout the course of writing out documentation.

PR #2, opened 2021-04-07
Docsp 13855 usage example find a doc

biniona-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13855

### Docs staging link (requires sign-in on MongoDB Corp SSO):
https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-13855-Usage-Example-Find-A-Doc/usage-examples/findOne/

### Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST?
- [x] Did you run a spell-check?
- [x] Did you run a grammar-check?
- [x] Does it render on staging correctly?
- [x] Are all the links working?
- [x] Are the staging links in the PR description updated?

PR #4, opened 2021-04-26
Docsp 13827

terakilobyte:
## Pull Request Info

This adds the quick-start for go

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13827

https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-13827/quick-start/
### Self-Review Checklist


cbush (approved):

terakilobyte:
@kevinAlbs ready for a pass from the team

source/includes/quick-start/main.go
```diff
@@ -0,0 +1,58 @@
...
+	"time"
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+var (
+	uri         = os.Getenv("MONGODB_URI")
+	ctx, cancel = context.WithTimeout(context.Background(), 20*time.Second)
```
iwysiu, line 18: ctx should be initialized in main()

source/includes/quick-start/main.go
```diff
@@ -0,0 +1,58 @@
...
+		if findResult.Err() == mongo.ErrNoDocuments {
+			fmt.Printf("No document was found with the title %s\n", title)
+			return
+		} else {
+			log.Panic(findResult.Err().Error())
+		}
+	}
+
+	var doc bson.D
+	if err = findResult.Decode(&doc); err != nil {
```
iwysiu, line 49: This can be absorbed into the above section, with the Decode call instead of findResult.Err(), since Decode will return findResult.Err() if it's non-nil

source/includes/quick-start/main.go
```diff
@@ -0,0 +1,58 @@
...
+		log.Panic(err)
+	}
+
+	defer func() {
+		if err = client.Disconnect(ctx); err != nil {
+			log.Panic(err)
+		}
+	}()
+
+	db := client.Database("sample_mflix")
```
iwysiu, line 34: nit: you can call `coll := client.Database("sample_mflix").Collection("movies")` as one line

iwysiu (commented):
Nice work! A couple comments.

source/includes/quick-start/main.go
terakilobyte: Ok, sounds good. I'll move it down.

terakilobyte:
@iwysiu I've made the changes you suggested, please have another look. Also, feel free to "request changes"!

source/includes/quick-start/main.go
```diff
@@ -0,0 +1,55 @@
...
+
+	findResult := coll.FindOne(ctx, bson.D{{"title", title}})
+
+	var doc bson.D
+	if err = findResult.Decode(&doc); err != nil {
+		if err == mongo.ErrNoDocuments {
+			fmt.Printf("No document was found with the title %s\n", title)
+			return
+		} else {
+			log.Panic(findResult.Err().Error())
```
iwysiu, line 46:
```suggestion
			log.Panic(err.Error())
```

iwysiu (approved):
lgtm, with one more nit

source/includes/quick-start/main.go
terakilobyte: Oh my, copy and paste error!

source/includes/quick-start/main.go
```diff
@@ -0,0 +1,55 @@
...
+	findResult := coll.FindOne(ctx, bson.D{{"title", title}})
+
+	var doc bson.D
+	if err = findResult.Decode(&doc); err != nil {
+		if err == mongo.ErrNoDocuments {
+			fmt.Printf("No document was found with the title %s\n", title)
+			return
+		} else {
+			log.Panic(err.Error())
+		}
```
iwysiu, line 47: You don't actually need the else, because the if statement returns
```suggestion
		}
		log.Panic(err.Error())
```

iwysiu (commented):
Sorry for coming back with one more change, I should be done now.

source/includes/quick-start/main.go
terakilobyte: No problem, and good pointing this out, thanks!

source/includes/quick-start/query-output.rst
```diff
@@ -0,0 +1,25 @@
+When you run ``main.go``, it should output the details of the
+movie from the sample dataset which will look something like this:
+
+.. code-block:: json
+
+   [
```
divjotarora, line 6: This is related to my comment about using `bson.D` in the `main.go` code. Because of the reasons explained there, this JSON is an array and doesn't actually reflect the document that's stored in the DB. When switching to `bson.M` in the code, please also update this file with the new JSON.

source/quick-start.txt
```diff
@@ -10,3 +10,68 @@ Quick Start
...
+
+.. code-block:: shell
+
+   mkdir go-quickstart
+   cd go-quickstart
+   go mod init go-quickstart
+
+.. _add-mongodb-dependency:
+
+Add MongoDB as a Dependency
```
divjotarora, line 31: Are we assuming previous Go knowledge on the part of the user? If not, we should explicitly state that this will download the latest released version of the Go Driver and that the user can instead add a different version by doing `go get go.mongodb.org/mongo-driver@<version tag>`.

divjotarora (changes requested):
This is looking great! Just jumping in with some suggestions.

source/quick-start.txt
```diff
@@ -10,3 +10,68 @@ Quick Start
...
+Query Your MongoDB Cluster from Your Application
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Next, create a file to contain your application called ``main.go``
+in the base directory of your project. Use the following sample
+code to run a query on your sample dataset in MongoDB Atlas.
+
+Set the value of the ``uri`` variable with your MongoDB Atlas connection
+string, or create an environmental variable with the name ``MONGODB_URI``
+with your Atlas connection string.
```
matthewdale, line 55: Consider adding an example shell command for setting the `MONGODB_URI` variable.

E.g.
```bash
export MONGODB_URI='<connection string>'
```

terakilobyte:
@divjotarora @matthewdale

Thanks for the feedback. I've incorporated the suggested changes. Ready for another look.

matthewdale (approved):
Looks good, thanks for the updates 👍

source/includes/quick-start/main.go
```diff
@@ -0,0 +1,53 @@
...
+	"os"
+	"time"
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+var (
+	uri = os.Getenv("MONGODB_URI")
```
kevinAlbs, line 17: Minor, but this is inconsistent with how the URI is passed in the findOne example, which is directly included in the source:
https://github.com/mongodb/docs-golang/pull/2/files#diff-9080c13327b8e149406ba2a276b31c02c0e664eb9db2a5a6c8c348f2a9a037ffR13

I can see an argument for either way. Passing through an environment variable is safer, and will probably save some new users from accidentally committing credentials. But for a quick getting started guide, it is one less step if the uri is directly in the source. Either way, it may be good to be consistent.

If you decide to stick with an environment variable, I'd suggest adding a check in `main` and panicking if it is unset. Currently, if you forget to set the environment variable `MONGODB_URI`, you'll get the following error, which might be unclear to users:

```
$ go run main.go
2021/05/07 11:58:56 error parsing uri: scheme must be "mongodb" or "mongodb+srv"
panic: error parsing uri: scheme must be "mongodb" or "mongodb+srv"
```

source/includes/quick-start/main.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
kevinAlbs, line 21: For consistency with the findOne example (https://github.com/mongodb/docs-golang/pull/2/files#diff-9080c13327b8e149406ba2a276b31c02c0e664eb9db2a5a6c8c348f2a9a037ffR16), it may be better to use `context.TODO()` as a placeholder context.

source/includes/quick-start/query-output.rst
```diff
@@ -0,0 +1,15 @@
+When you run ``main.go``, it should output the details of the
```
kevinAlbs, line 1: If we are assuming users are relatively new to Go, I suggest adding an explicit step for running `main.go` on the CLI:

```
go run main.go
```

Similar to the Node docs.

kevinAlbs (commented):
Looking good! I mostly have comments about consistency with the findOne docs.

source/includes/quick-start/main.go
```diff
@@ -0,0 +1,53 @@
...
+	err = coll.FindOne(ctx, bson.D{{"title", title}}).Decode(&result)
+	if err == mongo.ErrNoDocuments {
+		fmt.Printf("No document was found with the title %s\n", title)
+		return
+	}
+	if err != nil {
+		log.Panic(err)
+	}
+
+	jsonData, err := json.MarshalIndent(result, "", "    ")
```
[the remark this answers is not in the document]
terakilobyte: Quite alright. I think once we have a page detailing the differences between the two types and when to use them, perhaps bson.M for returned items from MongoDB and bson.D for sending (just spitballing here, likely quite different!) it'll be an improvement all around.

source/includes/quick-start/main.go
terakilobyte: I had thought about this but initially dismissed it as perhaps being too detailed. I'll run my idea by you in Slack.

source/quick-start.txt
```diff
@@ -67,6 +67,12 @@ with your Atlas connection string.
    :language: go
    :dedent:
 
+Run the sample code with teh following command from your command line:
```
kevinAlbs, line 70:
```suggestion
Run the sample code with the following command from your command line:
```

source/includes/quick-start/main.go
```diff
@@ -0,0 +1,54 @@
...
+		log.Panic(err)
+	}
+
+	defer func() {
+		if err = client.Disconnect(ctx); err != nil {
+			log.Panic(err)
+		}
+	}()
+
+	coll := client.Database("sample_mflix").Collection("movies")
```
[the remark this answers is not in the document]
kevinAlbs: @terakilobyte it might be worth adding the note blurb from the findOne page (https://github.com/mongodb/docs-golang/pull/2/files#diff-5e814de2f029ecad71b3ce40d12e14dcc1319d1f8d55d45efb21d2212ed230b6R1) to call out to users that they need to load the sample data

source/quick-start.txt
kevinAlbs: ping @terakilobyte

kevinAlbs (commented):
Looking good!

source/quick-start.txt
terakilobyte: We aren't assuming expert previous knowledge, but a minimum requirement would be knowing how to install Go and having read about the different built in tools. 

Since this is a quick start, I'm hesitant to add information about installing specific versions. If they are new I worry that adding information about installing specific versions can lead them to research it for very little gain and adds to the perceived complexity.

Definitely open to a discussion about this!

source/includes/quick-start/main.go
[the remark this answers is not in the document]
terakilobyte: I did add wording and a link to the load sample dataset step in the setup instructions, but adding another note may help users who weren't paying attention.

terakilobyte:
@kevinAlbs Thanks for the comments. The typo is fixed, and I've added another note about the sample datasets. I also replied to the comment about installing specific versions.

source/includes/quick-start/main.go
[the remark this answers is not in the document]
kevinAlbs: Thanks! I had not realized I suggested a self link :joy:

kevinAlbs (approved):
LGTM!

PR #5, opened 2021-07-22
DOCSP-13859: UpdateOne usage example

rustagir:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13859

### Docs staging link (requires sign-in on MongoDB Corp SSO):
https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-13859-update-one-usage-ex/usage-examples/updateOne/

**TODO:** edit links to Fundamentals pages "Change a Document" and "Retrieve Data" when they are created

### Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST?
- [x] Did you run a spell-check?
- [x] Did you run a grammar-check?
- [x] Does it render on staging correctly?
- [x] Are all the links working?
- [x] Are the staging links in the PR description updated?

PR #8, opened 2021-07-26
DOCSP 13856: find usage example

rustagir:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13856

https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-13856-find-usage-ex/usage-examples/find/

### Self-Review Checklist

### DIFF
--- a/source/code-snippets/find.go
+++ b/source/code-snippets/find.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+func main() {
+	// Replace the uri string with your MongoDB deployment's connection string.
+	uri := "mongodb+srv://<username>:<password>@<cluster-address>/test?w=majority"
+
+	ctx := context.TODO()
+	client, err := mongo.Connect(ctx, options.Client().ApplyURI(uri))
+
+	if err != nil {
+		panic(err)
+	}
+	defer func() {
+		if err = client.Disconnect(ctx); err != nil {
+			panic(err)
+		}
+	}()
+
+	// begin find
+	myCollection := client.Database("sample_training").Collection("zips")
+	filter := bson.D{{"pop", bson.D{{"$lte", 500}}}}
+
+	cur, err := myCollection.Find(ctx, filter)
+	// end find
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var results []bson.M
+	if err = cur.All(context.TODO(), &results); err != nil {
+		log.Fatal(err)
+	}
+	for _, result := range results {
+		fmt.Println(result)
+	}
+
+}
--- a/source/code-snippets/find.go
+++ b/source/code-snippets/find.go
@@ -14,14 +14,13 @@ func main() {
 	// Replace the uri string with your MongoDB deployment's connection string.
 	uri := "mongodb+srv://<username>:<password>@<cluster-address>/test?w=majority"
 
-	ctx := context.TODO()
-	client, err := mongo.Connect(ctx, options.Client().ApplyURI(uri))
+	client, err := mongo.Connect(context.TODO(), options.Client().ApplyURI(uri))
 
 	if err != nil {
 		panic(err)
 	}
 	defer func() {
-		if err = client.Disconnect(ctx); err != nil {
+		if err = client.Disconnect(context.TODO()); err != nil {
 			panic(err)
 		}
 	}()
@@ -30,7 +29,7 @@ func main() {
 	myCollection := client.Database("sample_training").Collection("zips")
 	filter := bson.D{{"pop", bson.D{{"$lte", 500}}}}
 
-	cur, err := myCollection.Find(ctx, filter)
+	cursor, err := myCollection.Find(context.TODO(), filter)
 	// end find
 
 	if err != nil {
@@ -38,7 +37,7 @@ func main() {
 	}
 
 	var results []bson.M
-	if err = cur.All(context.TODO(), &results); err != nil {
+	if err = cursor.All(context.TODO(), &results); err != nil {
 		log.Fatal(err)
 	}
 	for _, result := range results {
--- a/source/code-snippets/find.go
+++ b/source/code-snippets/find.go
@@ -9,10 +9,10 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
-func main() {
-	// Replace the uri string with your MongoDB deployment's connection string.
-	uri := "mongodb+srv://<username>:<password>@<cluster-address>/test?w=majority"
+// Replace the uri string with your MongoDB deployment's connection string.
+const uri = "mongodb+srv://<username>:<password>@<cluster-address>/test?w=majority"
 
+func main() {
 	client, err := mongo.Connect(context.TODO(), options.Client().ApplyURI(uri))
 
 	if err != nil {
--- a/source/code-snippets/find.go
+++ b/source/code-snippets/find.go
@@ -35,7 +35,7 @@ func main() {
 		panic(err)
 	}
 
-	var results []bson.M
+	var results []bson.D
 	if err = cursor.All(context.TODO(), &results); err != nil {
 		panic(err)
 	}
--- a/source/code-snippets/find.go
+++ b/source/code-snippets/find.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"log"
 
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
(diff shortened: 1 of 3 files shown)


source/usage-examples/find.txt
```diff
@@ -3,3 +3,54 @@ Find Multiple Documents
 =======================
 
 .. default-domain:: mongodb
+
+You can find multiple documents in a collection by using the ``Find()``
+method.
+
+The following example passes a query filter to the ``Find()`` method,
+which matches documents in the ``zips`` collection if the value of their
+``pop`` field is less than or equal to *500*:
```
biniona-mongodb, line 12:
```suggestion
``pop`` field is less than or equal to 500:
```

Suggestion: I don't think we generally use italics for values.

source/usage-examples/find.txt
```diff
@@ -3,3 +3,54 @@ Find Multiple Documents
 =======================
 
 .. default-domain:: mongodb
+
+You can find multiple documents in a collection by using the ``Find()``
+method.
+
+The following example passes a query filter to the ``Find()`` method,
+which matches documents in the ``zips`` collection if the value of their
```
biniona-mongodb, line 11:
```suggestion
The following example passes a query filter to the ``Find()`` method
that matches documents in the ``zips`` collection if the value of their
```

Nit: I think this should be "that" as I think the contents of the [query is essential information in the context of the page](https://www.grammarly.com/blog/which-vs-that/) (it is necessary for understanding expected results section).

source/code-snippets/find.go
```diff
@@ -0,0 +1,48 @@
...
+
+	cur, err := myCollection.Find(ctx, filter)
+	// end find
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var results []bson.M
+	if err = cur.All(context.TODO(), &results); err != nil {
```
biniona-mongodb, line 41:
```suggestion
	if err = cur.All(ctx, &results); err != nil {
```

Question: Is there a reason a new context was created here?

source/code-snippets/find.go
```diff
@@ -0,0 +1,48 @@
...
+		if err = client.Disconnect(ctx); err != nil {
+			panic(err)
+		}
+	}()
+
+	// begin find
+	myCollection := client.Database("sample_training").Collection("zips")
+	filter := bson.D{{"pop", bson.D{{"$lte", 500}}}}
+
+	cur, err := myCollection.Find(ctx, filter)
```
biniona-mongodb, line 33:
```suggestion
	cursor, err := myCollection.Find(ctx, filter)
```

Suggestion: Could this be changed to "cursor"? I think the ambiguity created by "cur" (could also mean current) is more harmful to readers than the additional three characters.

source/usage-examples/find.txt
```diff
@@ -3,3 +3,54 @@ Find Multiple Documents
...
+   :language: go
+   :dedent:
+
+Click here **<TODO: github link to file>** to see a fully runnable example.
+
+Expected Result
+---------------
+
+This example returns a ``Cursor`` object that contains
+documents similar to these:
```
biniona-mongodb, line 29:
```suggestion
This example creates a ``Cursor`` object that returns
documents similar to these:
```

Issue: I don't believe the above code snippet has a return value, so I think it is incorrect to say that something is returned.

Issue: I do not believe a cursor contains the result documents, but rather acts as an iterable that can retrieve them in batches.

source/usage-examples/find.txt
```diff
@@ -3,3 +3,54 @@ Find Multiple Documents
...
+
+.. code-block:: json
+   :copyable: false
+
+   // results truncated
+   ...
+   { ... , "city" : "SHAKTOOLIK", "zip" : "99771", ... , "pop" : 183, "state" : "AK" },
+   { ... , "city" : "DARLINGTON", "zip" : "83231", ... , "pop" : 12, "state" : "ID" },
+   { ... , "city" : "EAST BARRE", "zip" : "05649", ... , "pop" : 381, "state" : "VT" },
```
biniona-mongodb, line 38:
```suggestion
   { ...  "pop" : 183, "state" : "AK" },
   { ...  "pop" : 12, "state" : "ID" },
   { ...  "pop" : 381, "state" : "VT" },
```

Suggestion: Could the return value be shortened? I think the important thing is the population.

Nit: I think having two sets of ellipses in each sample return value is hard to read.

source/usage-examples/find.txt
biniona-mongodb, line 11:
```suggestion
which matches documents in the ``zips`` collection where the value of the
```

Suggestion: I think "matches documents if..." sounds slightly off, as it sounds like the database only goes through the act of matching if a certain condition is met (it always goes through the act of matching, rather some things do not fit the criteria it is looking for).

Suggestion: "their ``pop`` field" sounded strange to me, as it made me think on first reading that the group of documents returned shared a ``pop`` field.

biniona-mongodb (changes requested):
Awesome job. Looks great. After these should be good to go through.

source/usage-examples/find.txt
rustagir: Nathan initially suggested this formatting on my UpdateOne PR (#5). I thought I would keep it across usage example pages for consistency.

source/usage-examples/find.txt
rustagir: Leaving it as is to maintain consistency with other pages. Since there is so little information preceding the code snippet, I don't think people will skip that part of the sentence even if it starts with a "which"

source/code-snippets/find.go
rustagir: Updating all context statements to `context.TODO()` per https://github.com/mongodb/docs-golang/pull/5#discussion_r676969118.

source/code-snippets/find.go
rustagir: accepted

source/usage-examples/find.txt
rustagir: accepted

source/usage-examples/find.txt
rustagir: To allow readers to locate these documents to make sure the code worked, I think the city, state pair is necessary to disambiguate. Since the city names are also the same number of chars, all the fields are pretty much aligned anyways, making the results easy to visually process in my opinion. Omitted the zip field to eliminate horizontal scroll.

source/usage-examples/find.txt
rustagir: accepted

source/usage-examples/find.txt
biniona-mongodb: Sorry wasn't aware of this new convention. In that case, sounds good!

source/usage-examples/find.txt
biniona-mongodb: I don't think in this case it is worth emulating other examples, as the project is still really new (not much precedent). I think this change is inline with the [usage example template](https://docs.google.com/document/d/1hCXhDzcIXo3oqs9OzH158v98lMwIaJFkReln1y-TXjw/edit#), so I think it is better to get the wording as correct as possible rather than try and make the content as close to other content as possible.

source/usage-examples/find.txt
```diff
@@ -3,3 +3,54 @@ Find Multiple Documents
...
+documents similar to these:
+
+.. code-block:: json
+   :copyable: false
+
+   // results truncated
+   ...
+   { ... , "city" : "SHAKTOOLIK", ... , "pop" : 183, "state" : "AK" },
+   { ... , "city" : "DARLINGTON", ... , "pop" : 12, "state" : "ID" },
+   { ... , "city" : "EAST BARRE", ... , "pop" : 381, "state" : "VT" },
```
biniona-mongodb, line 38:
```suggestion
   { ... , "city" : "SHAKTOOLIK", "pop" : 183, "state" : "AK" },
   { ... , "city" : "DARLINGTON", "pop" : 12, "state" : "ID" },
   { ... , "city" : "EAST BARRE", "pop" : 381, "state" : "VT" },
```

[As documents are unordered](https://docs.mongodb.com/manual/reference/operator/aggregation/sort/#sort-consistency), I think it is unnecessary (and potentially implies an order in documents) to include two sets of ellipses.

biniona-mongodb (changes requested):
Just this one item and resolve the one item from last discussion and then good to go.

source/usage-examples/find.txt
biniona-mongodb: Why is city state pair the best pair for disambiguating these items? Does City and Population (or state and population) have a collision?

source/usage-examples/find.txt
biniona-mongodb: I was 100% wrong! Misread the sorts document. [Key/Values are ordered](https://bsonspec.org/spec.html), documents are unordered. Rea was right! Great job!

biniona-mongodb (approved):
LGTM!

rustagir:
Hi @kevinAlbs, this PR is ready for review.

source/code-snippets/find.go
```diff
@@ -0,0 +1,46 @@
...
+	filter := bson.D{{"pop", bson.D{{"$lte", 500}}}}
+
+	cursor, err := myCollection.Find(context.TODO(), filter)
+	// end find
+
+	if err != nil {
+		panic(err)
+	}
+
+	var results []bson.M
```
benjirewis, line 38: We normally avoid `bson.M` for regular use-cases since it's unordered and use `bson.D` instead. 
```suggestion
	var results []bson.D
```

benjirewis (approved):

PR #10, opened 2021-07-29
DOCSP 13853: bson fundies

rustagir:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13853

https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-13853-marshal-fundies/fundamentals/bson/

### Self-Review Checklist


source/fundamentals/bson.txt
```diff
@@ -0,0 +1,231 @@
...
+#. The Go Driver marshals embedded struct fields as subdocuments.
+   Each key is the lowercase of the field's type.
+
+#. The Go Driver marshals a pointer field as the underlying type if the
+   pointer is non-nil. If the pointer is nil, the driver marshals it as a BSON null
+   value.
+
+#. When unmarshalling, the Go Driver follows `these D/M type mappings
+   <https://pkg.go.dev/go.mongodb.org/mongo-driver/bson#hdr-Native_Go_Types>`_
+   for fields of type interface{}. The driver unmarshals BSON documents
```
terakilobyte, line 112:
```suggestion
   for fields of type ``interface{}``. The driver unmarshals BSON documents
```

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,231 @@
...
+   unmarshalled into an interface{} field as a ``D`` type.
```
terakilobyte, line 113:
```suggestion
   unmarshalled into an ``interface{}`` field as a ``D`` type.
```

terakilobyte (approved):
LGTM with two small formatting changes

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,231 @@
...
+-------------
+
+You can unmarshal BSON documents using the ``Decode()`` method. When you
+execute either a ``FindOne()`` or ``Find()`` query, you can use the
+``Decode()`` method on the return object to unmarshal the results.
+
+The ``Decode()`` method returns an ``error`` interface type which
+contains one of the following values:
+
+- ``nil`` if a document matched your query, and there were no errors retrieving
```
isabelatkinson, line 190:
```suggestion
- ``nil`` if a document matched your query, and there were no errors retrieving and unmarshalling
```

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,231 @@
...
+   * - ``minsize``
+     - If the field type is type int64, uint, uint32, or uint64 and the value of
+       the field can fit in a signed int32, the field will be serialized
+       as a BSON int32 rather than a BSON int64.
+
+   * - ``truncate``
+     - If the field type is a non-float numeric type, BSON doubles
+       unmarshalled into that field will be trucated at the decimal point.
+
+   * - ``inline``
```
isabelatkinson, line 91: Struct flattening can be a bit confusing sometimes for users so I think it'd be useful to have a concrete example of how values will roundtrip from struct to BSON back to struct when using this tag -- the serde documentation has a nice page about this: https://serde.rs/attr-flatten.html

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,231 @@
...
+   :header-rows: 1
+
+   * - Struct Tag
+     - Description
+
+   * - ``omitempty``
+     - The field will not be marshalled if it is set to the zero value
+       corresponding to the field type.
+
+   * - ``minsize``
```
isabelatkinson, line 82: It might be useful to specify what happens when this struct tag is used and the value cannot fit into a signed int32 (i.e. will an error occur?)

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,231 @@
...
+types to/from BSON using a `configurable codec system <https://pkg.go.dev/go.mongodb.org/mongo-driver/bson/bsoncodec>`_.
+
+For more information on how the Go Driver handles BSON data, see the
+`bson package API documentation
+<https://pkg.go.dev/go.mongodb.org/mongo-driver/bson>`_.
+
+Struct Tags
+-----------
+
+You can modify the default marshalling behavior of the Go Driver using
```
isabelatkinson, line 64:
```suggestion
You can modify the default marshalling and unmarshalling behavior of the Go Driver using
```

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,231 @@
...
+
+When used on the ``SingleResult`` type returned by the ``FindOne()``
+method, ``Decode()`` can also return the ``ErrNoDocuments`` error if no
+documents matched the query filter.
+
+The following example demonstrates how you can use the ``Decode()``
+method to unmarshal and read the result of a simple ``FindOne()``
+operation:
+
+.. code-block:: go
```
isabelatkinson, line 206: Would it be more informative to users here to show unmarshalling into a user-defined type (i.e. the `Student` type defined above) rather than `bson.D`? I'm not super familiar with the Go API, but in Rust, we return the `Document` type from our `find_one` method, so the Serde API is mostly useful to users when they're deserializing results back into their custom types.

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,231 @@
...
+   
+   myCollection := client.Database("school").Collection("students")
+   filter := bson.D{{"age", 8}}
+
+   var result bson.D
+   cursor := myCollection.FindOne(context.TODO(), filter)
+   err = cursor.Decode(&result)
+   ...
+
+   fmt.Println(result)
```
isabelatkinson, line 216: It might be useful to show what the output of this print will be to demonstrate what the data returned from `FindOne` was unmarshalled into.

isabelatkinson (commented):
looks good! just a few comments/suggestions

source/fundamentals/bson.txt
rustagir: added this to the example

source/fundamentals/bson.txt
rustagir: addressed

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,301 @@
...
+You should read this guide if you want to learn more about how the Go Driver
+represents BSON data or need to adjust default marshalling
+and unmarshalling behaviors.
+
+Data Types
+----------
+
+MongoDB stores documents in a binary representation called :manual:`BSON
+</reference/bson-types/>` that allows for easy and flexible data processing. The Go
+Driver represents BSON data using two families of types: ``D`` and ``Raw``.
```
matthewdale, line 29: The description of "two families of types" doesn't really make sense in the context of how the `D`/`E`/`M`/`A` and `Raw` types are used.

This should be something like:
```
The Go Driver provides four main types for working with BSON data:

- ``D``: An ordered representation of a BSON document
- ``E``: Represents a single element in a BSON document
- ``M``: An unordered representation of a BSON document
- ``A``: An ordered representation of a BSON array
```

The `Raw` type is mostly relevant to unmarshaling and probably doesn't need to be described here.

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,301 @@
...
+
+The ``D`` family contains four types:
+
+- ``D``: An ordered representation of a BSON document (slice)
+- ``M``: An unordered representation of a BSON document (map)
+- ``A``: An ordered representation of a BSON array
+- ``E``: A single element inside a D type
+
+Of these, you can use the ``D`` and ``M`` types to
+build representations of BSON using native Go types.
```
matthewdale, line 39: This sentence doesn't really make sense because you can use all of the BSON primitive types to define BSON docs in Go.

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,301 @@
...
+``bson.D`` type to match documents with a ``quantity`` field value greater
+than 100:
+
+.. code-block:: go
+   
+   filter := bson.D{{"quantity", bson.D{{"$gt", 100}}}}
+
+You can use ``Raw`` types to validate and retrieve elements from a slice
+of bytes, which is useful if you don't want to unmarshal BSON into
+another type.
```
matthewdale, line 51: The `Raw` type is mostly relevant to unmarshaling (specifically querying a document _without_ unmarshaling), so consider moving that information under the _Unmarshalling_ heading.

Consider a description like:
```
You can also use the ``Raw`` type to retrieve elements from a BSON document byte slice without
unmarshaling it to a Go value. This can be useful if you need to quickly lookup individual elements
without unmarshaling the entire BSON document.
```

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,301 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn about how the Go Driver handles conversions
+between BSON and Go objects. The process of converting a Go object to
+BSON is called **marshalling**, while the reverse process is called **unmarshalling**.
```
matthewdale, line 18: nit: The conventional spelling in Go docs always uses one "L" in these words. Consider changing the spelling to always use one "L".

E.g.
* marshaling
* marshaler
* unmarshaling
* unmarshaler

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,301 @@
...
+      - Sets the lowercase of the struct fields as the BSON field names
+      - Includes an empty ``lastname`` field
+      - Stores the ``Address`` field as a nested object
+
+Unmarshalling
+-------------
+
+You can unmarshal BSON documents using the ``Decode()`` method. When you
+execute either a ``FindOne()`` or ``Find()`` query, you can use the
+``Decode()`` method on the return object to unmarshal the results.
```
matthewdale, line 223: This sentence is a bit confusing. You can only call `Decode()` directly on the result of `FindOne()`; `Find()` returns `(*mongo.Cursor, error)`, so you can't call `Decode()` directly on the result. Also, you can call `Decode()` on any cursor, not just cursors returned by `Find()` (e.g. `Aggregate()` also returns a cursor).

Consider updating the description to say you can call `Decode()` on the result of `FindOne()` or any `*mongo.Cursor`.

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,301 @@
...
+
+The ``Decode()`` method returns an ``error`` interface type which
+contains one of the following values:
```
matthewdale, line 226: nit: "interface type" is redundant.

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,301 @@
...
+   
+   filter := bson.D{{"quantity", bson.D{{"$gt", 100}}}}
+
+You can use ``Raw`` types to validate and retrieve elements from a slice
+of bytes, which is useful if you don't want to unmarshal BSON into
+another type. 
+
+In Go, a **struct** is a collection of data fields with declared data
+types. The Go Driver can marshal/unmarshal structs and other native Go
+types to/from BSON using a `configurable codec system <https://pkg.go.dev/go.mongodb.org/mongo-driver/bson/bsoncodec>`_.
```
matthewdale, line 55: It's kinda weird to go straight into talking about structs, and then have a separate section for _Struct Tags_. Consider renaming the _Struct Tags_ section to something like _BSON as a Go `struct`_ and move this paragraph into the beginning of that section.

source/fundamentals/bson.txt
rustagir: got it. omitted info about Raw types and changed the list intro to your suggestion!

source/fundamentals/bson.txt
rustagir: ok-- omitted sentence

source/fundamentals/bson.txt
rustagir: added as a note to the end of the page

source/fundamentals/bson.txt
rustagir: Hmm the API Documentation seems to use two L's: https://pkg.go.dev/go.mongodb.org/mongo-driver@v1.7.0/bson#hdr-Marshalling_and_Unmarshalling

source/fundamentals/bson.txt
rustagir: I moved the sentence but kept the subheading of the section as Struct Tags because I think that's helpful for users navigating the "On this page" table of contents

source/fundamentals/bson.txt
```diff
@@ -0,0 +1,298 @@
...
+operation:
+
+.. code-block:: go
+   
+   myCollection := client.Database("school").Collection("students")
+   filter := bson.D{{"age", 8}}
+
+   var result bson.D
+   cursor := myCollection.FindOne(context.TODO(), filter)
+   err = cursor.Decode(&result)
```
matthewdale, line 242: The type returned by `FindOne` isn't a cursor (it returns a `*mongo.SingleResult`), so the variable naming here is confusing. Combine these two lines into the more common usage of `FindOne`:

```go
err := myCollection.FindOne(context.TODO(), filter).Decode(&result)
```

source/fundamentals/bson.txt
matthewdale: Ok, seems reasonable.

matthewdale (approved):
Looks good 👍

isabelatkinson (approved):

PR #42, opened 2021-09-02
DOCSP-18444 example section

kyuan-mongodb:
## Pull Request Info

Updated all example sections to the following:
- Contain an Example heading
- Move the tip box before the example statement starts

The example statements have the following formats: (omitting the parameters and method name now)
- For a list of actions: `“The following example performs ...”` 
- For non-list actions: `“The following example matches ...”`
- For insert pages: `“The following example inserts ...”`
- For watch: `“The following example opens ...”`
- For command: `“The following example retrieves ...”`
- For struct-tagging: `“The following example creates ...”`

NOTE: Only the watch page shows how to print because that showcases how to best use it

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-18444

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=61390773c608bec57f8cd9e6

https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-18444-ExampleSection/usage-examples/

### Self-Review Checklist

- [x] Are the staging and workerpool job links in the PR description updated?

### If your page documents a concept, does it meet the following criteria?

- [ ] Target the [Jasmin persona](https://drive.google.com/file/d/14FbBOLCVxwSP6M9BK4Nz1Ir9tzxT8_02/view)
- [ ] Target the [Lucas persona](https://drive.google.com/file/d/1J2vqJxo7ldv7OP_obA9Q-avf0o_ju4Lk/view)

### DIFF
--- a/source/usage-examples/bulkWrite.txt
+++ b/source/usage-examples/bulkWrite.txt
@@ -16,7 +16,7 @@ The following example performs the following in order on the ``haikus``
 collection:
 
 - Matches a document in which the ``title`` is "Record of a Shriveled Datum" and replace it with a new document
-- Matches a document in which the ``title`` is "Dodging Greys" and update that field to "Dodge The Greys"
+- Matches a document in which the ``title`` is "Dodging Greys" and updates that value to "Dodge The Greys"
 
 .. literalinclude:: /includes/usage-examples/code-snippets/bulk.go
    :start-after: begin bulk
--- a/source/usage-examples/bulkWrite.txt
+++ b/source/usage-examples/bulkWrite.txt
@@ -12,8 +12,8 @@ Example
 
 .. include:: /includes/usage-examples/run-example-tip.rst
 
-The following example passes a slice of ``WriteModels`` and an option
-to the ``BulkWrite()`` method, which performs the following in order:
+The following example performs the following in order on the ``haikus``
+collection:
 
 - Match a document where the ``title`` is "Record of a Shriveled Datum" and replace it with a new document
 - Match a document where the ``title`` is "Dodging Greys" and update that field to "Dodge The Greys"
--- a/source/usage-examples/bulkWrite.txt
+++ b/source/usage-examples/bulkWrite.txt
@@ -7,13 +7,16 @@ Perform Bulk Operations
 You can perform bulk write operations on a collection by using the
 ``BulkWrite()`` method.
 
-The following example specifies a slice of ``WriteModels`` and an option
+Example
+-------
+
+.. include:: /includes/usage-examples/run-example-tip.rst
+
+The following example passes a slice of ``WriteModels`` and an option
 to the ``BulkWrite()`` method, which performs the following in order:
 
 - Match a document where the ``title`` is "Record of a Shriveled Datum" and replace it with a new document
 - Match a document where the ``title`` is "Dodging Greys" and update that field to "Dodge The Greys"
-
-.. include:: /includes/usage-examples/run-example-tip.rst
 
 .. literalinclude:: /includes/usage-examples/code-snippets/bulk.go
    :start-after: begin bulk
--- a/source/usage-examples/bulkWrite.txt
+++ b/source/usage-examples/bulkWrite.txt
@@ -15,8 +15,8 @@ Example
 The following example performs the following in order on the ``haikus``
 collection:
 
-- Match a document where the ``title`` is "Record of a Shriveled Datum" and replace it with a new document
-- Match a document where the ``title`` is "Dodging Greys" and update that field to "Dodge The Greys"
+- Matches a document in which the ``title`` is "Record of a Shriveled Datum" and replace it with a new document
+- Matches a document in which the ``title`` is "Dodging Greys" and update that field to "Dodge The Greys"
 
 .. literalinclude:: /includes/usage-examples/code-snippets/bulk.go
    :start-after: begin bulk
--- a/source/usage-examples/count.txt
+++ b/source/usage-examples/count.txt
@@ -14,11 +14,11 @@ Example
 
 .. include:: /includes/usage-examples/run-example-tip.rst
 
-The following example uses the ``EstimatedDocumentCount()`` method to
-count the number of documents in the ``movies`` collection. Then, the
-example passes a query filter to the ``CountDocuments()`` method, which
-counts the number of documents in the ``movies`` collection where the
-``countries`` field contains "China".
+The following example performs the following on the ``movies``
+collection:
+
+- Approximately count the number of documents in the collection
+- Count the number of documents where the ``countries`` field contains "China"
 
 .. literalinclude:: /includes/usage-examples/code-snippets/count.go
    :start-after: begin countDocuments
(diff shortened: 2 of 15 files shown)


source/usage-examples/count.txt
```diff
@@ -9,14 +9,17 @@ collection by using the ``EstimatedDocumentCount()`` method and an exact
...
-example uses the ``CountDocuments()`` method to count the number of
-documents in the ``movies`` collection where the ``countries`` field
-contains "China".
+Example
+-------
 
 .. include:: /includes/usage-examples/run-example-tip.rst
 
+The following example uses the ``EstimatedDocumentCount()`` method to
+count the number of documents in the ``movies`` collection. Then, the
```
ccho-mongodb, line 18: Personally, I'm not sure the user or writer gets any value from standardizing the structure of this sentence for all usage examples, but I think this example and a couple others do not match the one in your PR description:

> Updated all example sections to the following:
> The example statement with the format: “The following example passes .... to the ___ method, which ...”

Applies to some of the updated instances.

source/usage-examples/count.txt
```diff
@@ -9,14 +9,17 @@ collection by using the ``EstimatedDocumentCount()`` method and an exact
...
+example passes a query filter to the ``CountDocuments()`` method, which
+counts the number of documents in the ``movies`` collection where the
+``countries`` field contains "China".
```
ccho-mongodb, line 21: This seems like a lot of information for a single sentence. I wonder if you might consider using a bulleted list or at least separate sentences instead?

Applies to all instances.

ccho-mongodb (changes requested):
A couple items:
1) Some changes don't seem to match with the format described in the PR.
2) Some of the sentences are difficult to read because they contain multiple subjects, causes, and effects. Not sure if this was a group decision, but I think if it was, you might reconsider and ask others for additional opinions onthis if necessary.

source/usage-examples/deleteMany.txt
```diff
@@ -7,13 +7,15 @@ Delete Multiple Documents
...
-method, which matches documents in the ``movies`` collection where the
-``runtime`` field is greater than *800* and deletes all the matching
-documents. 
+Example
+-------
 
 .. include:: /includes/usage-examples/run-example-tip.rst
 
+The following example matches documents in the ``movies`` collection
+where the ``runtime`` field is greater than 800, deleting all
```
ccho-mongodb, line 16: Question: are we ok with referring to field names as the value they contain? Or should we just say:
"where the runtime is greater than 800 [minutes]"?

ccho-mongodb (changes requested):
Question: Is the description in this PR no longer accurate in which you wanted to make the wording conform to:
“The following example passes .... to the ___ method, which ...”?

The current format I see is "The following example performs/retrieves/matches", no mention of the method, and some inconsistency in the wording of the effect (e.g. "and prints..." vs ", deleting/returning ..."

It's completely normal to change the intent of a PR to something else, but could you make sure to let your reviewer(s) know?

source/usage-examples/deleteMany.txt
kyuan-mongodb: Confirmed that we should include the unit of measure in which it is applicable.

ccho-mongodb: Ah ok, I wasn't so concerned about the unit of measure which is why I included it in square brackets. The mainpoint here is disambiguating "field" from "value".

source/usage-examples/deleteMany.txt
```diff
@@ -7,13 +7,15 @@ Delete Multiple Documents
...
-method, which matches documents in the ``movies`` collection where the
-``runtime`` field is greater than *800* and deletes all the matching
-documents. 
+Example
+-------
 
 .. include:: /includes/usage-examples/run-example-tip.rst
 
+The following example matches documents in the ``movies`` collection
+in which the ``runtime`` field is greater than 800 min, deleting all
```
ccho-mongodb, line 16: Either use "minutes" or omit the unit of measurement. I think "min" is more confusing that either alternative since it could stand for "minimum".

source/usage-examples/updateMany.txt
```diff
@@ -7,14 +7,17 @@ Update Multiple Documents
...
-documents by a factor of *1.15*:
+Example
+-------
 
 .. include:: /includes/usage-examples/run-example-tip.rst
 
+The following example performs the following on the
+``listingsAndReviews`` collection:
+
+- Matches documents in which the ``address.market`` is "Sydney"
```
ccho-mongodb, line 18: I wonder whether we should expect our usage example readers to know dot notation?
Maybe you could either:
- include a note about what dot notation is and link to more info (perhaps https://docs.mongodb.com/manual/core/document/#dot-notation)
- explain this is the notation they can use to search the market field of the address subdocument

source/usage-examples/updateMany.txt
```diff
@@ -7,14 +7,17 @@ Update Multiple Documents
...
+- Multiplies the ``price`` in the matched documents by 1.15
```
ccho-mongodb, line 19: I think it could help the clarity of this bullet point to mention that it updates the document with the new prices.
Otherwise, it could be interpreted as just performing a multiplication operation and maybe printing that out instead of making any updates to the matched document(s).

source/usage-examples/struct-tagging.txt
```diff
@@ -7,10 +7,15 @@ Use Struct Tags
...
-field name:
+Example
+-------
+
+.. include:: /includes/usage-examples/run-example-tip.rst
+
+The following code declares a ``BlogPost`` struct, in which the
+``WordCount`` has a struct tag that sets a custom BSON field name
+of ``word_count``. By default, the driver marshals the other fields as
+the lowercase of the struct field name:
```
ccho-mongodb, line 18: This sentence is a little long and might be easier to read if separated.
"The following code declares a struct of type BlogPost. This struct contains a struct tag that maps the "WordCount" field to the BSON field name word_count. By default, the driver marshals "

source/usage-examples/bulkWrite.txt
```diff
@@ -7,14 +7,17 @@ Perform Bulk Operations
...
+Example
+-------
 
 .. include:: /includes/usage-examples/run-example-tip.rst
 
+The following example performs the following in order on the ``haikus``
+collection:
+
+- Matches a document in which the ``title`` is "Record of a Shriveled Datum" and replace it with a new document
+- Matches a document in which the ``title`` is "Dodging Greys" and update that field to "Dodge The Greys"
```
ccho-mongodb, line 19: A couple items:
- "update" needs to have verb agreement
- it's not really the field, but rather a value attached to a field.

```suggestion
- Matches a document in which the ``title`` is "Dodging Greys" and updates that value to "Dodge The Greys"
```

ccho-mongodb (approved):
LGTM. I added a few suggestions that I think could help improve it though.

source/usage-examples/updateMany.txt
kyuan-mongodb: I'll update this to the second bullet

PR #70, opened 2021-10-14
DOCSP-19088 code snippet updates

kyuan-mongodb:
## Pull Request Info

Updated all code snippets to the following:

- Use `coll` for collection
- Use `err` for error (except the count page since it shows two examples together)
- Use `opts` for options
- Use `cursor` for results from the `Find()` function
- Use `results` to store cursor values
- Use `filter` for all filters
- Show the error checking for `err` because a user should know right away if there's an error

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-19088

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=6168853c8670bd2731828e0c

https://docs-mongodborg-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-19088-CodeSnippetUpdates/
(see the files tab for changes in each file)

### Self-Review Checklist

- [x] Are the staging and workerpool job links in the PR description updated?

### If your page documents a concept, does it meet the following criteria?

- [ ] Target the [Jasmin persona](https://drive.google.com/file/d/14FbBOLCVxwSP6M9BK4Nz1Ir9tzxT8_02/view)
- [ ] Target the [Lucas persona](https://drive.google.com/file/d/1J2vqJxo7ldv7OP_obA9Q-avf0o_ju4Lk/view)

### DIFF
--- a/source/includes/usage-examples/code-snippets/findOne.go
+++ b/source/includes/usage-examples/code-snippets/findOne.go
@@ -20,7 +20,7 @@ func main() {
 
 	var uri string
 	if uri = os.Getenv("MONGODB_URI"); uri == "" {
-		log.Fatal("You must set your 'MONGODB_URI' environmental variable. See\n\t https://docs.mongodb.com/drivers/go/current/usage-examples/")
+		log.Fatal("You must set your 'MONGODB_URI' environmental variable. See\n\t https://docs.mongodb.com/drivers/go/current/usage-examples/#environment-variable")
 	}
 
 	client, err := mongo.Connect(context.TODO(), options.Client().ApplyURI(uri))
--- a/source/includes/usage-examples/code-snippets/struct-tag.go
+++ b/source/includes/usage-examples/code-snippets/struct-tag.go
@@ -32,7 +32,7 @@ func main() {
 
 	var uri string
 	if uri = os.Getenv("MONGODB_URI"); uri == "" {
-		log.Fatal("You must set your 'MONGODB_URI' environmental variable. See\n\t https://docs.mongodb.com/drivers/go/current/usage-examples/")
+		log.Fatal("You must set your 'MONGODB_URI' environmental variable. See\n\t https://docs.mongodb.com/drivers/go/current/usage-examples/#environment-variable")
 	}
 
 	client, err := mongo.Connect(context.TODO(), options.Client().ApplyURI(uri))
--- a/source/fundamentals/crud/read-operations/sort.txt
+++ b/source/fundamentals/crud/read-operations/sort.txt
@@ -74,7 +74,6 @@ The following example specifies an ascending sort on the ``rating`` field:
    :dedent:
    :start-after: begin ascending sort
    :end-before: end ascending sort
-   :emphasize-lines: 2-3, 5
 
 After running this example, the output resembles the following:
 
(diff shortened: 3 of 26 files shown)


source/includes/usage-examples/code-snippets/struct-tag.go
```diff
@@ -32,7 +32,7 @@ func main() {
 
 	var uri string
 	if uri = os.Getenv("MONGODB_URI"); uri == "" {
-		log.Fatal("You must set your `MONGODB_URI' environmental variable. See\n\t https://docs.mongodb.com/drivers/go/current/usage-examples/")
+		log.Fatal("You must set your 'MONGODB_URI' environmental variable. See\n\t https://docs.mongodb.com/drivers/go/current/usage-examples/")
```
ccho-mongodb, line 35: Question:
Should this URL be
https://docs.mongodb.com/drivers/go/current/usage-examples/#environment-variable instead so that the reader doesn't have to scan through the page to find it?

source/includes/usage-examples/code-snippets/findOne.go
```diff
@@ -35,16 +35,18 @@ func main() {
...
 	// begin findOne
 	coll := client.Database("sample_mflix").Collection("movies")
+	
 	var result bson.M
 	err = coll.FindOne(context.TODO(), bson.D{{"title", "The Room"}}).Decode(&result)
-	// end findOne
 	if err != nil {
 		if err == mongo.ErrNoDocuments {
 			// This error means your query did not match any documents.
 			return
```
ccho-mongodb, line 44: The code sample currently prints a result whether a runtime error or a result, and nothing if it doesn't match any documents. Would it be helpful to print the query didn't match any documents in this conditional block?

source/includes/usage-examples/code-snippets/findOne.go
```diff
@@ -35,16 +35,18 @@ func main() {
 
 	// begin findOne
 	coll := client.Database("sample_mflix").Collection("movies")
+
```
ccho-mongodb, line 38: Adding a line here broke your emphasize-lines setting. I wasn't able to check many of these since it's not that apparent which code snippets correspond to which page in all cases. Could you check that they all render correctly as expected, and fix the ones that aren't quite right like this one?

ccho-mongodb (approved):
Looks good. There's one type of issue you should fix, but I trust you'll be able to make the appropriate changes without me having to re-review. Also left a couple questions.

source/includes/usage-examples/code-snippets/struct-tag.go
kyuan-mongodb: Good catch! I agree that the URL should be updated.

source/includes/usage-examples/code-snippets/findOne.go
kyuan-mongodb: The example in the API Docs for [FindOne()](https://pkg.go.dev/go.mongodb.org/mongo-driver@v1.7.2/mongo#Collection.FindOne) only returns, like this example currently does. I noticed that when I create code snippets that differ from the API examples, the engineers usually ask me to fix mine to match.

PR #109, opened 2022-05-25
DOCSP-22623 output block updates

kyuan-mongodb:
# Pull Request Info

Updated all output blocks to not be visible and to have its language as `none`.

JIRA - <https://jira.mongodb.org/browse/DOCSP-22623>
Staging - <https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-22623/>

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/aggregation.txt
+++ b/source/fundamentals/aggregation.txt
@@ -144,7 +144,8 @@ the ``$sum`` expression operator.
       }
 
    .. output::
-      :language: go
+      :language: none
+      :visible: false
 
       Average price of black tea: $6.075 
       Amount of black tea: 4 
@@ -197,7 +198,8 @@ The aggregation pipleline contains the following stages:
       }
 
    .. output::
-      :language: go
+      :language: none
+      :visible: false
 
       Tea: Hojicha 
       Toppings: [lemon ginger milk foam] 
--- a/source/fundamentals/bson.txt
+++ b/source/fundamentals/bson.txt
@@ -254,6 +254,7 @@ operation:
 
    .. output::
       :language: none
+      :visible: false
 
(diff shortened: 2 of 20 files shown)


ccho-mongodb (approved):
LGTM
I'm not 100% convinced that we're helping the reader by collapsing every fundamentals example that shows output, so not sure that part of the PR is useful, but don't have a strong opinion. I'm curious to hear what reasons you have for it though.

kyuan-mongodb:
So the reason I decided to collapse the output for the fundamentals pages is because I didn't think that every reader would want to see the output. I typically only look for the code and disregard/skim past the output on sites I visit, and assumed there are a handful of folks like me. So by collapsing it, it'll be available for those that do want to see it, but won't have readers that don't care to see it potentially skip over relevant content on the page. 

I am open to leaving them open if you and the rest of the team think its better to leave it in.

ccho-mongodb:
> So the reason I decided to collapse the output for the fundamentals pages is because I didn't think that every reader would want to see the output. I typically only look for the code and disregard/skim past the output on sites I visit, and assumed there are a handful of folks like me. So by collapsing it, it'll be available for those that do want to see it, but won't have readers that don't care to see it potentially skip over relevant content on the page.
> 
> I am open to leaving them open if you and the rest of the team think its better to leave it in.

Ok, that's fine. As mentioned before, I don't have a strong opinion. It can be an extra click or something they can't cmd/ctrl-F for on the page nor print, but the tradeoff is that they might be using the page as you described and want to ignore the output.

Let's just make sure we're mindful about whether we're helping or detracting from the user's experience for each change rather than as a blanket application.

PR #175, opened 2022-10-27
(DOCSP-26180) Bump Min Go Version

biniona-mongodb:
# Pull Request Info

Bumps min go version

JIRA - <https://jira.mongodb.org/browse/DOCSP-26180>

[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-26180/compatibility/#language-compatibility)

## Self-Review Checklist

### DIFF
--- a/source/includes/language-compatibility-table-go.rst
+++ b/source/includes/language-compatibility-table-go.rst
@@ -1,1 +1,46 @@
-The MongoDB Go Driver requires Go {+min-go-version+} or later.
+The following compatibility table specifies the recommended versions of the
+{+driver-long+} for use with a specific version of the Go programming language.
+
+The first column lists the driver version.
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+   :class: compatibility-large
+
+   * - Go Driver Version
+     - Go 1.13
+     - Go 1.10
+   * - 1.10
+     - ✓
+     - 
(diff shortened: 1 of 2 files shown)


source/includes/language-compatibility-table-go.rst
```diff
@@ -1 +1,34 @@
...
+   * - 1.4
+     - 1.10
+   * - 1.3
+     - 1.10
+   * - 1.2
+     - 1.10
+   * - 1.1
+     - 1.10
+   * - 1.0
+     - 1.10
```
rustagir, line 34: Why did you choose to go with versions in the right column instead of checkmarks? I think this presentation might be a little confusing, for example, if users see this table and at first glance think they can only use Golang v1.10 for Driver v1.9. Since the version numbers are also similar between Golang and the driver, it might also reduce confusion to have a matrix.

biniona-mongodb: Discussed sync, agreed it this makes sense as `Go` only has a minimum language version, where as other drivers can have minimum and maximum language versions.

rustagir (approved):
LGTM

benjirewis (approved):
LGTM! Thanks @biniona-mongodb 🧑‍🔧

PR #176, opened 2022-10-28
DOCSP-18237: run a command

rustagir:
# Pull Request Info

[PR Reviewing Guidelines](https://github.com/mongodb/docs-golang/blob/master/REVIEWING.md)

JIRA - https://jira.mongodb.org/browse/DOCSP-18237
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-18237-runCommand/fundamentals/crud/run-command/

## Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST?
- [x] Did you run a spell-check?
- [x] Did you run a grammar-check?
- [x] Are all the links working?

PR #178, opened 2022-10-31
DOCSP-18237 run a command Backport #176 to v1.10

rustagir:
Ports https://github.com/mongodb/docs-golang/pull/176 to v1.10.

Staging - https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/v1.10/fundamentals/crud/run-command/

### DIFF
--- a/source/fundamentals/crud/run-command.txt
+++ b/source/fundamentals/crud/run-command.txt
@@ -1,0 +1,167 @@
+.. _golang-run-command:
+
+=============
+Run a Command
+=============
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to run a **database command** with the
+{+driver-short+}. You can use database commands to perform a variety of
+administrative and diagnostic tasks, such as fetching server statistics,
+initializing a replica set, or running an aggregation pipeline.
+
+Compose a Command
+-----------------
+
+The {+driver-short+} provides the following methods to run database
+commands: 
+
+- ``RunCommand()``, which returns the command response as a
+  ``SingleResult`` type. You can use this method with any database command.
+- ``RunCommandCursor()``, which returns the command response as a
+  ``Cursor`` type. You can use this method if your database command
+  returns multiple result documents.
+
+Each method requires a Context and a command document. The
+command document must be an order-preserving type such as ``bson.D``.
+The following code shows how you can use the ``RunCommandCursor()``
+method to run the ``listCollections`` command on a database:
+
+.. code-block:: go
+
+   command := bson.D{{"listCollections", 1}}   
+   cursor, err := db.RunCommandCursor(context.TODO(), command)
+
+For a full list of database commands, see the :ref:`Additional
+Information <addl-info-runcommand>`.
+
+.. note:: Read Preference
+
+   ``RunCommand()`` and ``RunCommandCursor()`` do not obey the read
+   preference set for the database. You can set a read preference for a
+   database command by passing a ``RunCmdOptions`` type to either method:
+   
+   .. code-block:: go
+
+      opts := options.RunCmd().SetReadPreference(readpref.Primary())
+      err := db.RunCommand(context.TODO(), command, opts).Decode(&result)
+   
+   For more information on
+   read preference options, see the :ref:`golang-write-read-pref`
+   fundamentals page.
+
+Response
+--------
+
+Each method returns a document or a cursor containing documents with the
+following fields:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 70
+
+   * - Field
+     - Description
+
+   * - <command result>
+     - Provides fields specific to the database command. For example,
+       ``count`` returns the ``n`` field and ``explain`` returns the
+       ``queryPlanner`` field.
+
+   * - ``ok``
+     - Indicates whether the command has succeeded (``1``)
+       or failed (``0``).
+
+   * - ``operationTime``
+     - Indicates the :website:`logical time
+       </blog/post/transactions-background-part-4-the-global-logical-clock>`
+       of the operation. MongoDB uses the logical time to order operations.
+
+   * - ``$clusterTime``
+     - Provides a document that returns the signed cluster time. Cluster time is a
+       logical time used for ordering of operations.
+
+       The document contains the following fields:
+
+       - ``clusterTime``, which is the timestamp of the highest known cluster time for the member.
+       - ``signature``, which is a document that contains the hash of the cluster time and the ID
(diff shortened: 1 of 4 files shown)


source/fundamentals/crud/run-command.txt
```diff
@@ -0,0 +1,167 @@
...
+In this guide, you can learn how to run a **database command** with the
+{+driver-short+}. You can use database commands to perform a variety of
+administrative and diagnostic tasks, such as fetching server statistics,
+initializing a replica set, or running an aggregation pipeline.
+
+Compose a Command
+-----------------
+
+The {+driver-short+} provides the following methods to run database
+commands:
```
ccho-mongodb, line 27: A couple items related to this section:

1) The title is "compose a command", but I wasn't able to find information about what it means to compose a command.  
2) The list below shows methods that you can use to wrap the commands, but lacks explanation of what it means to run and compose a database command.
3) Is there any info you can provide to explain when a user might care about using this command? The example shows listCollections for which you can usually use [ListCollections()](https://pkg.go.dev/go.mongodb.org/mongo-driver@v1.10.3/mongo#Database.ListCollections), so it's not self-evident how these wrapper methods are helpful.

source/fundamentals/crud/run-command.txt
```diff
@@ -0,0 +1,167 @@
...
+-----------------
+
+The {+driver-short+} provides the following methods to run database
+commands: 
+
+- ``RunCommand()``, which returns the command response as a
+  ``SingleResult`` type. You can use this method with any database command.
+- ``RunCommandCursor()``, which returns the command response as a
+  ``Cursor`` type. You can use this method if your database command
+  returns multiple result documents.
```
ccho-mongodb, line 33:
```suggestion
  ``Cursor`` type. You must use this method to access multiple result documents if your database command
  returns multiple result documents.
```

source/fundamentals/crud/run-command.txt
```diff
@@ -0,0 +1,167 @@
...
+
+Each method requires a Context and a command document. The
+command document must be an order-preserving type such as ``bson.D``.
```
ccho-mongodb, line 36: Issue: command document hasn't been defined, so it can be confusing to the reader to only read about the type requirements rather than what it is. I think it could be helpful to explain the parts of the command document since it isn't self-evident in the example (e.g. what's the "1" value used for?)

source/fundamentals/crud/run-command.txt
```diff
@@ -0,0 +1,167 @@
...
+   command := bson.D{{"listCollections", 1}}   
+   cursor, err := db.RunCommandCursor(context.TODO(), command)
+
+For a full list of database commands, see the :ref:`Additional
+Information <addl-info-runcommand>`.
+
+.. note:: Read Preference
+
+   ``RunCommand()`` and ``RunCommandCursor()`` do not obey the read
+   preference set for the database. You can set a read preference for a
```
ccho-mongodb, line 51: Suggestion: clarify "for the database". I think this needs to describe the Database object they created, so maybe something like "... do not obey the read preference you set on your Database object" (not sure of the exact terminology that you've decided to use for this driver).

source/fundamentals/crud/run-command.txt
```diff
@@ -0,0 +1,167 @@
...
+   database command by passing a ``RunCmdOptions`` type to either method:
```
ccho-mongodb, line 52: Issue: I think you meant "object" or "instance" rather than type. A type only refers to a data format it corresponds with.

source/fundamentals/crud/run-command.txt
```diff
@@ -0,0 +1,167 @@
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 70
+
+   * - Field
+     - Description
+
+   * - <command result>
+     - Provides fields specific to the database command. For example,
```
ccho-mongodb, line 77: Issue: I think this could be more clear if it explained that it contained the results the database returned after running the command. 

Alternatively, explain this before the table since it doesn't currently specify the conditions in which it's returned.

source/fundamentals/crud/run-command.txt
```diff
@@ -0,0 +1,167 @@
...
+       ``count`` returns the ``n`` field and ``explain`` returns the
+       ``queryPlanner`` field.
+
+   * - ``ok``
+     - Indicates whether the command has succeeded (``1``)
+       or failed (``0``).
+
+   * - ``operationTime``
+     - Indicates the :website:`logical time
+       </blog/post/transactions-background-part-4-the-global-logical-clock>`
```
ccho-mongodb, line 87: Issue: avoid linking to longer blog articles to explain terminology. Also avoid trying to teach terminology in setting descriptions when possible. If it's not possible to avoid, use a seealso admonition to create a separation between the description and additional learning. In this case, it seems like the Server Manual should have information, given that it is a concept the Server uses and calls out, and that reading a blog post might not be the right type of learning they're looking for here.

source/fundamentals/crud/run-command.txt
```diff
@@ -0,0 +1,167 @@
...
+      err := db.RunCommand(context.TODO(), command, opts).Decode(&result)
+   
+   For more information on
+   read preference options, see the :ref:`golang-write-read-pref`
+   fundamentals page.
+
+Response
+--------
+
+Each method returns a document or a cursor containing documents with the
```
ccho-mongodb, line 66: In many cases, I like the idea of making it convenient to reference everything on a single page. From doing some research, it looks like these fields are the same as https://github.com/10gen/docs-mongodb-internal/blob/master/source/reference/method/db.runCommand.txt

I haven't checked how the Go driver handles RunCommand, but if it is just returning what the manual page references, then it might be best to direct the reader to the server documentation relevant to the version of the server they're connecting to. That would make maintenance easier for us in case it differs by version.

source/fundamentals/crud/run-command.txt
```diff
@@ -0,0 +1,167 @@
...
+       logical time used for ordering of operations.
+
+       The document contains the following fields:
+
+       - ``clusterTime``, which is the timestamp of the highest known cluster time for the member.
+       - ``signature``, which is a document that contains the hash of the cluster time and the ID
+         of the key used to sign the cluster time.
+
+Example
+-------
```
ccho-mongodb, line 101: For reasons mentioned earlier, I think it's odd to use an example for which we'd prefer they use the driver method for counting documents if there aren't any benefits.

source/fundamentals/crud/run-command.txt
rustagir: You can still use `RunCommand` if the command returns multiple documents.

source/fundamentals/crud/run-command.txt
rustagir: the server manual does not have an explanation, but I have used a `seealso` admonition.

PR #186, opened 2022-11-04
DOCSP-13830: enterprise auth mechanisms

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-13830
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-13830-enterprise-auth/fundamentals/enterprise-auth/

## Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?

### DIFF
--- a/source/fundamentals/auth.txt
+++ b/source/fundamentals/auth.txt
@@ -20,8 +20,10 @@ mechanism** available in the MongoDB Community Edition. MongoDB uses
 authentication mechanisms to confirm an identity and establish trust to
 ensure security in the driver and server before connecting.
 
-If you want to establish a connection to your MongoDB cluster without an
-authentication mechanism, see :ref:`golang-connection-guide`.
+To authenticate using ``GSSAPI/Kerberos`` or ``LDAP``, see the
+:ref:`golang-enterprise-authentication-mechanisms` fundamentals page. For more 
+information on establishing a connection to your MongoDB cluster, see
+the :ref:`golang-connection-guide`.
 
 Supported Mechanisms
 --------------------
--- a/source/fundamentals/enterprise-auth.txt
+++ b/source/fundamentals/enterprise-auth.txt
@@ -1,27 +1,184 @@
+.. _golang-enterprise-authentication-mechanisms:
+
 ====================================
 Enterprise Authentication Mechanisms
 ====================================
 
-.. default-domain:: mongodb
-
 .. contents:: On this page
    :local:
    :backlinks: none
-   :depth: 1
+   :depth: 2
    :class: singlecol
 
-In this guide, we show you how to authenticate with MongoDB using each
-**authentication mechanism** available exclusively in the MongoDB Enterprise
-Edition.
+Overview
+--------
 
-You can use the following mechanisms with the latest version of the MongoDB
-Enterprise Edition:
+In this guide, you can learn how to authenticate with MongoDB using the
+**authentication mechanisms** available in the MongoDB Enterprise Edition.
+Authentication mechanisms are processes by which the driver and server confirm
+the identity of a client to ensure security before connecting.
 
-Kerberos (GSSAPI)
-LDAP (PLAIN)
+You can use the following authentication mechanisms with the latest version of
+the MongoDB Enterprise Edition.
 
-To authenticate using another mechanism, see
-:ref:`golang-authentication-mechanisms`. To learn more about 
-establishing a connection to your MongoDB cluster, see 
+- :ref:`GSSAPI/Kerberos <golang-kerberos>`
+- :ref:`LDAP (Plain) <golang-LDAP>`
+
+To authenticate using another mechanism, see the
+:ref:`<golang-authentication-mechanisms>` fundamentals page. For
+more information on establishing a connection to your MongoDB cluster, see the
 :ref:`golang-connection-guide`.
 
+You can specify your authentication mechanism and credentials when connecting to
+MongoDB using the following methods:
+
+- Using a **connection string**, which is a string
+  that tells the driver how to connect to a MongoDB deployment and how to behave while
+  connected.
+
+- Specifying credentials and an authentication mechanism in the
+  ``Credential`` type from the {+driver-short+}. You can then use your
+  ``Credential`` struct to provide options for the ``Client``.
+
+This guide provides information on using types and methods provided by
+the {+driver-short+} for enterprise authentication, rather than using a
+connection string. For more information on using a connection string for
+enterprise authentication, see the :manual:`manual entry on Connection
+String URIs </reference/connection-string/#authentication-options>`.
+
+.. _golang-kerberos:
+
+Authenticate with GSSAPI/Kerberos
+---------------------------------
+
+The Generic Security Services API (GSSAPI) authentication mechanism allows the 
+user to authenticate to a Kerberos service using the user's principal name.
+
+The following example specifies the authentication mechanism using the following
+placeholders:
+
+- ``username``: Your :wikipedia:`URL-encoded <Percent-encoding>`
+  principal name. An sample username is "myuser@KERBEROS.EXAMPLE.COM"
+- ``password``: Your Kerberos user's password.
+- ``hostname``: The network address of your MongoDB server, accessible by your client.
+- ``port``: Your MongoDB server's port number.
+
+.. note::
+
+   Using Kerberos authentication requires the "gssapi" build tag and ``cgo``
+   support during compilation. The default service name for Kerberos is
+   "mongodb". You can define the ``AuthMechanismProperties`` field in
+   the ``Credential`` struct to use a different service name.
+
+The following code shows how you can define a ``Credential`` struct to
+authenticate to Kerberos and create a client with your authentication
+preferences:
+
+.. code-block:: go
+
+   credential := options.Credential{
+   		  AuthMechanism: "GSSAPI",
+   		  Username:      "<username>",
+           Password:      "<password>",
+   }
+   uri := "mongodb://<hostname>:<port>"
+   clientOpts := options.Client().ApplyURI(uri).SetAuth(credential)
+   
+   client, err := mongo.Connect(context.TODO(), clientOpts)
+
+.. tip::
+
+   You don't need to define a password in your ``Credential`` struct if
+   you store authentication keys in ``keytab`` files. You can initialize
+   a credential cache for authenticating the user principal using the
+   ``kinit`` binary. The following command shows how you can invoke a
+   credential cache for a sample username:
+   
+   .. code-block:: sh
+   
+      kinit myuser@KERBEROS.EXAMPLE.COM
+
+.. _golang-LDAP:
+
+Authenticate with LDAP (PLAIN)
+------------------------------
+
+You can authenticate to a Lightweight Directory Access Protocol (LDAP) server 
+using your directory-server username and password.
+
+.. warning::
+
+   This authentication mechanism sends the password to the server in
+   plaintext, so you should use this mechanism only with TLS connections.
+
+The following example specifies the authentication mechanism using the following
+placeholders:
+
+- ``username``: Your LDAP username
+- ``password``: Your LDAP password
+- ``hostname``: The network address of your MongoDB server, accessible
+  by your client
+- ``port``: Your MongoDB server's port number
+
+The following code shows how you can define a ``Credential`` struct to
+authenticate to LDAP and create a client with your authentication
+preferences:
+
+.. code-block:: go
+
+   credential := options.Credential{
+		     AuthMechanism: "PLAIN",
+		     Username:      "<username>",
+		     Password:      "<password>",
+	}
+   uri := "mongodb://<hostname>:<port>"
+   clientOpts := options.Client().ApplyURI(uri).SetAuth(credential)
+   
+   client, err := mongo.Connect(context.TODO(), clientOpts)
+
+.. tip::
+
+   The method refers to PLAIN instead of LDAP since it
+   authenticates using the PLAIN Simple Authentication and Security Layer
+   (SASL) defined in `RFC-4616 <https://tools.ietf.org/html/rfc4616>`__.
+         
+Configuration Options
+---------------------
+
+You can specify additional properties with your authentication
+mechanism using the ``AuthMechanismProperties`` field in the
+``Credential`` struct. The following code shows how you can set values
+for the ``SERVICE_NAME`` and ``SERVICE_REALM`` fields when defining a
+``Credential`` struct:
+
+.. code-block:: go
+
+   credential := options.Credential{
+		      AuthMechanism: "GSSAPI",
+		      Username:      "<username>",
+		      Password:      "<password>",
+		      AuthMechanismProperties: map[string]string{
+		      	     "SERVICE_REALM": "<Kerberos service realm>",
+                    "SERVICE_NAME":  "<service name>",
+		      },
+	}
+
+For a full list of properties, see the
+:ref:`manual entry on authentication properties </reference/connection-string/#mongodb-urioption-urioption.authMechanismProperties>`.
+
+Additional Information
+----------------------
+
+For more information about the concepts in this guide, see the following documentation:
+
+- :manual:`Kerberos Authentication </core/kerberos/>`
+- :manual:`LDAP Proxy Authentication </core/security-ldap/>`
+- :ref:`<golang-connection-guide>`
+- :ref:`<golang-authentication-mechanisms>`
+- :manual:`Connection String URI Format </reference/connection-string/>`
+
+API Documentation
+~~~~~~~~~~~~~~~~~
+
+- `Credential <{+api+}/mongo/options#Credential>`__
+- `SetAuth() <{+api+}/mongo/options#ClientOptions.SetAuth>`__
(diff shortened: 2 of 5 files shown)


source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,186 @@
...
+- ``username``: Your :wikipedia:`URL-encoded <Percent-encoding>`
+  principal name. A sample username is *myuser@KERBEROS.EXAMPLE.COM*.
+- ``password``: Your Kerberos user's password.
+- ``hostname``: The network address of your MongoDB server, accessible by your client.
+- ``port``: Your MongoDB server's port number.
+
+.. note::
+
+   Using Kerberos authentication requires the "gssapi" build tag and ``cgo``
+   support during compilation. The default service name for Kerberos is
```
ccho-mongodb, line 69: Suggestion: add an example or link to reference material on how a reader can add the build tag and cgo support.

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,186 @@
...
+
+This guide provides information on using types and methods provided by
+the {+driver-short+} for enterprise authentication, rather than using a
+connection string. For more information on using a connection string for
+enterprise authentication, see the :manual:`manual entry on Connection
+String URIs </reference/connection-string/#authentication-options>`.
+
+.. _golang-kerberos:
+
+Authenticate with GSSAPI/Kerberos
```
ccho-mongodb, line 51: Issue: I noticed three different prepositions used interchangeably for "authenticate" even though what it's modifying is one of two different things.

Suggestion: 
I think we should avoid "with" whenever possible since it is the least specific.

When describing the mechanism: "Authenticate using <auth mechanism>". 
When describing the server/srivce: "Authenticate to <server/service>".

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,186 @@
...
+---------------------------------
+
+The Generic Security Services API (GSSAPI) authentication mechanism allows the 
+user to authenticate to a Kerberos service using the user's principal name.
+
+The following example specifies the authentication mechanism using the following
```
ccho-mongodb, line 57: Issue: the example is introduced twice, so it was unclear where it begins. It's also unclear to me where it ends because rather than an example, I see placeholders and specifying the Credential struct.

Suggestion:
I think restructuring this section using the following ordering could be helpful:
1) What is this mechanism (already your first paragraph)
2) What are the requirements (build tag, compilation support, specifying credentials in Credential struct, alternatively keytab, etc)
3) Example

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,186 @@
...
-establishing a connection to your MongoDB cluster, see 
+To authenticate using another mechanism, see the
+:ref:`<golang-authentication-mechanisms>` fundamentals page. For
+more information on establishing a connection to your MongoDB cluster, see the
 :ref:`golang-connection-guide`.
 
+You can specify your authentication mechanism and credentials when connecting to
+MongoDB using the following methods:
+
+- Using a **connection string**, which is a string
```
ccho-mongodb, line 35: Suggestion: I think we refer to this as the Connection URI in the Connection Guide, so it could be more clear if used consistently. I think we could also defer the explanation to the Connection Guide.

Comment applies to all instances of "connection string".

ccho-mongodb (changes requested):
Great job in describing these auth mechanisms! I think there are some changes and
verbiage that we can update to make it easier for the reader to understand what we're showing. I also wanted to make sure to be picky with the style guide issues to build awareness.

source/fundamentals/auth.txt
```diff
@@ -20,8 +20,10 @@ mechanism** available in the MongoDB Community Edition. MongoDB uses
 authentication mechanisms to confirm an identity and establish trust to
 ensure security in the driver and server before connecting.
 
-If you want to establish a connection to your MongoDB cluster without an
-authentication mechanism, see :ref:`golang-connection-guide`.
```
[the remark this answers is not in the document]
rustagir: Not sure, I didn't write that page. I guess it's good that I removed it?

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,186 @@
...
+           },
+   }
+
+For a full list of properties, see the
+:manual:`manual entry on authentication properties </reference/connection-string/#mongodb-urioption-urioption.authMechanismProperties>`.
+
+Additional Information
+----------------------
+
+For more information about the concepts in this guide, see the following documentation:
```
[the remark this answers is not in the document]
rustagir: fixed in multiple places

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,186 @@
...
+- :manual:`Kerberos Authentication </core/kerberos/>`
+- :manual:`LDAP Proxy Authentication </core/security-ldap/>`
+- :ref:`<golang-connection-guide>`
+- :ref:`<golang-authentication-mechanisms>`
+- :manual:`Connection String URI Format </reference/connection-string/>`
+
+API Documentation
+~~~~~~~~~~~~~~~~~
+
+- `Credential <{+api+}/mongo/options#Credential>`__
```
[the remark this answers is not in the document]
rustagir: changed -- should I format this differently?

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,186 @@
...
+           Username:      "<username>",
+           Password:      "<password>",
+           AuthMechanismProperties: map[string]string{
+                   "SERVICE_REALM": "<Kerberos service realm>",
+                   "SERVICE_NAME":  "<service name>",
+           },
+   }
+
+For a full list of properties, see the
+:manual:`manual entry on authentication properties </reference/connection-string/#mongodb-urioption-urioption.authMechanismProperties>`.
```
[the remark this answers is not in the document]
rustagir: changed in both places

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,186 @@
...
+   
+   client, err := mongo.Connect(context.TODO(), clientOpts)
+
+.. tip::
+
+   The method refers to PLAIN instead of LDAP since it
+   authenticates using the PLAIN Simple Authentication and Security Layer
+   (SASL) defined in `RFC-4616 <https://tools.ietf.org/html/rfc4616>`__.
+         
+Configuration Options
```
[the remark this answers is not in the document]
rustagir: The only additional option that isn't for Kerberos is `AWS_SESSION_TOKEN`. I think I'll just move this content under the Kerberos header.

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,186 @@
...
+
+.. code-block:: go
+
+   credential := options.Credential{
+           AuthMechanism: "PLAIN",
+           Username:      "<username>",
+           Password:      "<password>",
+	}
+
+   uri := "mongodb://<hostname>:<port>"
```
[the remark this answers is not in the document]
rustagir: I think that they can use either any kind of connection string. By placeholder do you mean it should look like this?

```suggestion
   uri := ""
```

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,186 @@
...
+---------------------------------
+
+The Generic Security Services API (GSSAPI) authentication mechanism allows the 
+user to authenticate to a Kerberos service using the user's principal name.
+
+The following example specifies the authentication mechanism using the following
+placeholders:
+
+- ``username``: Your :wikipedia:`URL-encoded <Percent-encoding>`
+  principal name. A sample username is *myuser@KERBEROS.EXAMPLE.COM*.
```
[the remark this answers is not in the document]
rustagir: According to the [test examples](https://github.com/mongodb/mongo-go-driver/blob/8bcd4675d0e9b329df1b0899d431c54ef19ac265/mongo/client_examples_test.go#L230) it is user principal but I will double check. Also, I actually think usernames do not need to be URL encoded for the Go driver. removed that section.

source/fundamentals/enterprise-auth.txt
rustagir: I added a heading to delineate the example from the introductory text for each mechanism

source/fundamentals/enterprise-auth.txt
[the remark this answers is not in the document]
ccho-mongodb: I think method is self-evident when using "()" as mentioned in my comment, so not sure any change is needed to that line.

> should I format this differently?

I'm not sure what you're referring to here, could you explain further?

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,192 @@
...
+Additional Information
+----------------------
+
+To learn more about the concepts in this guide, see the following documentation:
+
+- :manual:`Kerberos Authentication </core/kerberos/>`
+- :manual:`LDAP Proxy Authentication </core/security-ldap/>`
+- :ref:`<golang-connection-guide>`
+- :ref:`<golang-authentication-mechanisms>`
+- :manual:`Connection String URI Format </reference/connection-string/>`
```
ccho-mongodb, line 186: Suggestion:

I think the Go driver docs is more relevant (unless you've linked to it already):
https://www.mongodb.com/docs/drivers/go/current/fundamentals/connection/

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,192 @@
...
-To authenticate using another mechanism, see
-:ref:`golang-authentication-mechanisms`. To learn more about 
-establishing a connection to your MongoDB cluster, see 
+To authenticate using another mechanism, see the
+:ref:`<golang-authentication-mechanisms>` fundamentals page. To learn
+more about establishing a connection to your MongoDB cluster, see the
 :ref:`golang-connection-guide`.
 
+You can specify your authentication mechanism and credentials when connecting to
+MongoDB using the following methods:
```
ccho-mongodb, line 33: Question: Is this general information or can you specify everything you need for GSSAPI and LDAP in a connection string?

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,192 @@
...
+
+- Using a **connection URI**. To learn more about using a connection
```
ccho-mongodb, line 35: Issue: the introduction to the list includes "using...", so starting this line with "using" is redundant.

Suggestion: "A **connection string URI**"

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,192 @@
...
+  URI for enterprise authentication, see the :manual:`Server manual
+  entry on Connection String URIs </reference/connection-string/#authentication-options>`.
```
ccho-mongodb, line 37: Suggestion:
Choose one form of the term and stick with it for the entire page (and at some point in the future, all the Go driver docs) whether "connection string", "connection URI", "connection string URI", etc.

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,192 @@
...
+.. _golang-kerberos:
+
+Authenticate to GSSAPI/Kerberos
+-------------------------------
+
+The Generic Security Services API (GSSAPI) authentication mechanism allows the 
+user to authenticate to a Kerberos service using the user's principal.
+
+You must use the "gssapi" `build tag
+<https://www.programming-books.io/essential/go/conditional-compilation-with-build-tags-d1980344374d45c082c914c2aafa50cf>`__
```
ccho-mongodb, line 56: Issue: I think it's best to use authoritative sources when possible. This appears to be a programming books site, which might not have reliable information nor stay at the same URL for long.

Suggestion: Use the official docs, perhaps this link: https://pkg.go.dev/go/build#hdr-Build_Constraints

source/fundamentals/enterprise-auth.txt
ccho-mongodb: The heading looks good. However, I think the restructuring is still necessary -- I left another related comment [here](https://github.com/mongodb/docs-golang/pull/186/files#r1016913399) with an additional explanation.

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,192 @@
...
+- ``username``: Your Kerberos principal. A sample username is ``myuser@KERBEROS.EXAMPLE.COM``.
+- ``password``: Your Kerberos user's password.
+- ``hostname``: The network address of your MongoDB server, accessible by your client.
+- ``port``: Your MongoDB server's port number.
+
+.. tip::
+
+   You don't need to define a password in your ``Credential`` struct if
+   you store authentication keys in ``keytab`` files. You can initialize
+   a credential cache for authenticating the user principal using the
```
ccho-mongodb, line 73: Question: should this be Kerberos principal per Benji's response, or is "user principal" the correct term here?

source/fundamentals/enterprise-auth.txt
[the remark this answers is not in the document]
ccho-mongodb: As mentioned on Slack, I meant:

```
<connection URI>
```

but based on the server manual section title, probably:

```
<connection string URI>
```

source/fundamentals/enterprise-auth.txt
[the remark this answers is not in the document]
ccho-mongodb: Suggestion: 
I would have expected to see the options prior to the example. Is there a good reason to place them after the example? If so, I think it could be helpful to introduce the subsections in the overview of the top-level section so that the reader can locate the information more quickly (rather than relying on the right-hand On this page nav).

source/fundamentals/enterprise-auth.txt
ccho-mongodb: Looks like you already linked to the connection guide. To make it what I think would be easier and more approachable would be to follow the style guide's recommendations on describing what they contain and to what site they're going.

source/fundamentals/enterprise-auth.txt
rustagir: you can authenticate with the enterprise methods using a connection string. https://www.mongodb.com/docs/manual/reference/connection-string/#authentication-options

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,192 @@
...
+You can specify your authentication mechanism and credentials when connecting to
+MongoDB using the following methods:
+
+- Using a **connection URI**. To learn more about using a connection
+  URI for enterprise authentication, see the :manual:`Server manual
+  entry on Connection String URIs </reference/connection-string/#authentication-options>`.
+
+- Specifying credentials and an authentication mechanism in the
+  ``Credential`` type from the {+driver-short+}. You can then use your
+  ``Credential`` struct to provide options for the ``Client``.
```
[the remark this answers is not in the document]
rustagir: See my response above

source/fundamentals/enterprise-auth.txt
rustagir: I saw the official documentation but I didn't feel like it laid out how to utilize build tags and then reference them when compiling.

rustagir: is that fine? I picked this other source because it had examples and more descriptive information.

source/fundamentals/enterprise-auth.txt
```diff
@@ -1,27 +1,192 @@
...
+
+The :ref:`example <golang-kerberos-example>` specifies the authentication mechanism using the following
+placeholders:
+
+- ``username``: Your Kerberos principal. A sample username is ``myuser@KERBEROS.EXAMPLE.COM``.
+- ``password``: Your Kerberos user's password.
+- ``hostname``: The network address of your MongoDB server, accessible by your client.
+- ``port``: Your MongoDB server's port number.
+
+.. tip::
```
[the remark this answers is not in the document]
rustagir: moved the placeholder descriptions under the example heading.

source/fundamentals/enterprise-auth.txt
```diff
@@ -93,19 +85,32 @@ preferences:
...
+   uri := "<connection uri>"
    clientOpts := options.Client().ApplyURI(uri).SetAuth(credential)
    
    client, err := mongo.Connect(context.TODO(), clientOpts)
 
-Configuration Options
-~~~~~~~~~~~~~~~~~~~~~
+You don't need to define a password or the ``PasswordSet`` field in
+your ``Credential`` struct if you store authentication keys in
+``keytab`` files. You can initialize a credential cache for
```
ccho-mongodb, line 98: Question: Could you explain why I might initialize a credential cache in the docs? Does it affect the driver behavior in any way? If so, it would be useful to document. If not, I think maybe it would be better to omit.

source/fundamentals/enterprise-auth.txt
ccho-mongodb: I see, that relates to a comment in my latest review. The linked content doesn't make it that easy to figure out what they need for the specific authentication option, so I think it might be beneficial to show examples of connection strings. What do you think of that idea?

source/fundamentals/enterprise-auth.txt
ccho-mongodb: We resolved this in a sync discussion.

ccho-mongodb (approved):
Just a few suggestions to improve the flow and provide some background info that could help the reader.

PR #204, opened 2022-11-29
DOCSP-25508: retry logic

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-25508
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-25508-retry-logic/fundamentals/connection/#single-timeout-setting (just the note)

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/connection.txt
+++ b/source/fundamentals/connection.txt
@@ -245,9 +245,9 @@ that Context deadline for the operation. If the context does not have a
 deadline, the driver derives a new Context from the given Context using
 the ``Timeout`` option set on the ``Client``.
 
-.. note::
+.. note:: Retries under Timeout Specification
 
-   If you set a ``Timeout`` option on your ``Client``
+   With default settings, if you set a ``Timeout`` option on your ``Client``
    and your operation requires a retry, the driver retries the operation
    as many times as possible before the timeout expires. Once the timeout
    expires, the driver returns a timeout error. Versions 1.1 and later
--- a/source/fundamentals/connection.txt
+++ b/source/fundamentals/connection.txt
@@ -245,15 +245,15 @@ that Context deadline for the operation. If the context does not have a
 deadline, the driver derives a new Context from the given Context using
 the ``Timeout`` option set on the ``Client``.
 
-.. note:: Retryability
+.. note::
 
-   This version of the {+driver-short+} enables retryable reads and
-   writes by default. If you set a ``Timeout`` option on your ``Client``
+   If you set a ``Timeout`` option on your ``Client``
    and your operation requires a retry, the driver retries the operation
    as many times as possible before the timeout expires. Once the timeout
-   expires, the driver returns a timeout error. See the Server manual
-   for more information about :manual:`retryable reads </core/retryable-reads/>` and
-   :manual:`retryable writes </core/retryable-writes/>`.
+   expires, the driver returns a timeout error. Versions 1.1 and later
+   of the {+driver-short+} enable retryable reads and writes by default.
+   See the Server manual for more information about :ref:`retryable
+   reads <retryable-reads>` and :ref:`retryable writes <retryable-writes>`.
 
 The following code shows how to set the ``Timeout`` option on a ``Client``
 with the ``SetTimeout`` option:
--- a/source/fundamentals/connection.txt
+++ b/source/fundamentals/connection.txt
@@ -253,7 +253,7 @@ the ``Timeout`` option set on the ``Client``.
    expires, the driver returns a timeout error. Versions 1.1 and later
    of the {+driver-short+} enable retryable reads and writes by default.
    See the Server manual for more information about :ref:`retryable
-   reads <retryable-reads>` and :ref:`retryable writes <retryable-writes>`.
+   reads <retryable-reads>` and :manual:`retryable writes </core/retryable-writes/>`.
 
 The following code shows how to set the ``Timeout`` option on a ``Client``
 with the ``SetTimeout`` option:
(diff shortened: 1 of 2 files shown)


source/fundamentals/connection.txt
```diff
@@ -245,6 +245,16 @@ that Context deadline for the operation. If the context does not have a
 deadline, the driver derives a new Context from the given Context using
 the ``Timeout`` option set on the ``Client``.
 
+.. note:: Retryability
+
+   This version of the {+driver-short+} enables retryable reads and
+   writes by default. If you set a ``Timeout`` option on your ``Client``
+   and your operation requires a retry, the driver retries the operation
+   as many times as possible before the timeout expires. Once the timeout
```
ccho-mongodb, line 253: Suggestion:
Mention that you can use the maxConnecting parameter to rate limit the new connection attempts if they observe the server getting flooded.

As a reader, when I saw "as many times as possible", I was concerned that this might flood a connection until I read "[Why don't drivers use backoff/jitter between retry attempts"](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.rst#why-dont-drivers-use-backoff-jitter-between-retry-attempts) in the spec.

If that detail doesn't seem important, then feel free not to mention anything. I'm also not certain whether maxConnecting is a Timeout option.

source/fundamentals/connection.txt
```diff
@@ -245,6 +245,16 @@ that Context deadline for the operation. If the context does not have a
 deadline, the driver derives a new Context from the given Context using
 the ``Timeout`` option set on the ``Client``.
 
+.. note:: Retryability
+
+   This version of the {+driver-short+} enables retryable reads and
```
ccho-mongodb, line 250: Issue:
This text will become inaccurate when releasing newer driver versions of the docs unless changed and changing details like this is generally outside of our normal version update process.

Suggestion:
Use "Starting in (version)...", or the versionadded directive.

source/fundamentals/connection.txt
```diff
@@ -245,6 +245,16 @@ that Context deadline for the operation. If the context does not have a
...
+   writes by default. If you set a ``Timeout`` option on your ``Client``
+   and your operation requires a retry, the driver retries the operation
+   as many times as possible before the timeout expires. Once the timeout
+   expires, the driver returns a timeout error. See the Server manual
+   for more information about :manual:`retryable reads </core/retryable-reads/>` and
```
ccho-mongodb, line 255: Suggestion:
Since the snooty file imports the refs from the Server Manual, you can use the retryable-reads and retryable-writes ref targets instead of the manual directive. These targets are hypothetically more durable than the url.

source/fundamentals/connection.txt
```diff
@@ -245,6 +245,16 @@ that Context deadline for the operation. If the context does not have a
 deadline, the driver derives a new Context from the given Context using
 the ``Timeout`` option set on the ``Client``.
 
+.. note:: Retryability
```
ccho-mongodb, line 248: Issue: 

I think the note title "Retryability" doesn't match the message and that the message introduces two different things, but should maybe only focus on one. 

To me, the term retryability describes how easily you can try to deliver a command or message that failed by some measure. Although the term is used as a header in the spec, I would avoid using that term to introduce this information.

The two separate topics I'm reading are:
1) "starting in (version), retryable reads and writes are enabled by default"
2) "When retryable reads/writes are enabled and the timeout option is set, there's a specific behavior"

Suggestion:
Since the reader may have already enabled retryable reads or writes, describe the conditions that lead to the retry behavior first. E.g.

"Starting in (version), when you enable retryable reads or writes and set the Timeout option in your Client, the driver (describe retry behavior)"

And then the final sentence (or possibly somewhere else) mention the other piece of information, e.g.:
"Retryable reads and writes are enabled by default starting in (version)"

ccho-mongodb (changes requested):
Looks pretty good; identified a couple issues.

source/fundamentals/connection.txt
rustagir: There is a Client option in the Go Driver to set `MaxConnecting`, but I'm not sure that mentioning this detail relates directly to the note. Setting `MaxConnecting` seems to be a safeguard when multiple concurrent operations select the same server. The spec doesn't seem to recommend setting `MaxConnecting` in general.

source/fundamentals/connection.txt
rustagir: The user does not have to enable retryable reads or writes, this behavior is enabled by default for nearly all versions of the Go driver. I removed the title and rearranged the sentences.

ccho-mongodb: Oh ok, that part wasn't clear to me (that the default values are enabled prior to this version).
I think it still might be important to mention the initial conditions since they might not be using the default settings for reads/writes, but will leave that up to you.

source/fundamentals/connection.txt
ccho-mongodb: I just thought it might help address an engineer's concern about connection floods/storms because retry as many times as fast as possible isn't an expected connection behavior. As mentioned in the linked doc, there's usually backoff in which it slows down the retries after each failure. Fine with omitting info about MaxConnecting.

source/fundamentals/connection.txt
```diff
@@ -245,6 +245,16 @@ that Context deadline for the operation. If the context does not have a
 deadline, the driver derives a new Context from the given Context using
 the ``Timeout`` option set on the ``Client``.
 
+.. note::
```
ccho-mongodb, line 248: Suggestion:
Given the structure of the admonition content, maybe the title could help surface what cases the content pertains to earlier:

```suggestion
.. note:: Retries with the Timeout Option Set
```

ccho-mongodb (approved):
LGTM. Added a suggestion, but am fine merging regardless!

PR #206, opened 2022-12-02
DOCSP-26830 - Clarify compound operations

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-26830
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/docsp-26830-find-and-update/fundamentals/crud/compound-operations/

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/crud/compound-operations.txt
+++ b/source/fundamentals/crud/compound-operations.txt
@@ -18,7 +18,7 @@ Overview
 In this guide, you can learn how to perform **compound operations**.
 
 Compound operations combine a read and write operation into
-a single operation.  If you perform a read and write operation
+a single operation. If you perform a read and write operation
 separately, there's a chance someone else may alter the document between
 both operations. MongoDB prevents this by placing a write lock on the
 document you are modifying for the duration of your compound operation.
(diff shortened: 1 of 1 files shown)


ccho-mongodb (changes requested):
Issue:

For both of these mentions, I think there's a gap in what the reader may be able to infer. I think this gap also exists in the current content. The FindOne + Update/DeleteOne equivalency needs more explanation -- it might not make immediate sense why the reader should perform each operation sequentially since UpdateOne() and DeleteOne() accomplish what they need.

I think mention of "atomicity" is important to explain the differentiation (currently explained as "eliminates the possibility of someone altering a document between both operations"). 

Suggestion:

I think listing the FindOne() and UpdateOne() methods works, but needs further qualification why someone might need to use those in succession and why you might need them to be atomic.
Since UpdateOne() is atomic, Benji mentioned avoiding the method since it could lead to some inference that it somehow performs multiple operations itself. But the only differences between UpdateOne() and FindOneAndUpdate() is that you can't provide a sort order (not considering collations) and return value (I think). See https://www.mongodb.com/docs/manual/reference/command/findAndModify/#comparisons-with-the-update-method for more info.

Without this qualification, I think the reader might get confused.
"If you need to find a document using a sort order and update it, use this method instead of FindOne() and UpdateOne(). This method performs the find and update atomically which means it executes the commands in a single operation. This eliminates..."

PR #219, opened 2022-12-15
DOCSP-26442: use struct in projection pg

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-26442
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-26442-struct-project/fundamentals/crud/read-operations/project/

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/crud/read-operations/project.txt
+++ b/source/fundamentals/crud/read-operations/project.txt
@@ -21,27 +21,44 @@ document from read operations.
 Sample Data
 ~~~~~~~~~~~
 
-To run the examples in this guide, load these documents into the
-``tea.ratings`` collection with the following
-snippet:
+The examples in this guide use the following ``Course`` struct as a model for documents
+in the ``courses`` collection:
 
-.. literalinclude:: /includes/fundamentals/code-snippets/CRUD/projection.go 
+.. literalinclude:: /includes/fundamentals/code-snippets/CRUD/projection.go
+   :start-after: start-course-struct
+   :end-before: end-course-struct
+   :language: go
+   :dedent:
+
+The ``omitempty`` :ref:`struct tag<golang-struct-tags>` directs the
+driver to exclude fields when unmarshalling based on your projection
+specification.
+
+To run the examples in this guide, load the sample data into the
+``db.courses`` collection with the following snippet:
+
+.. literalinclude:: /includes/fundamentals/code-snippets/CRUD/projection.go
    :language: go
    :dedent:
    :start-after: begin insertDocs
    :end-before: end insertDocs
 
-.. include:: /includes/fundamentals/tea-sample-data-ending.rst
+.. include:: /includes/fundamentals/automatic-db-coll-creation.rst
+
+Each document contains a description of a university course that
+includes the course title, course ID, and maximum enrollment, corresponding to
+the ``title``, ``course_id``, and ``enrollment`` fields in each document.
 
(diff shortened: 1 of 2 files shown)


source/fundamentals/crud/read-operations/project.txt
```diff
@@ -57,149 +74,132 @@ method in the options of the following read operation methods:
 Exclude a Field
 ~~~~~~~~~~~~~~~
 
-To exclude a field, pass the field you want to exclude and a ``0`` to
-the ``SetProjection()`` method. For all fields you don't explicitly
-list in the projection, the driver includes them.
+To exclude a field, pass the field you want to exclude with a ``0`` to
+the ``SetProjection()`` method. The driver includes all fields that are
+not explicitly excluded in the projection document.
```
DBirtolo-mdb, line 79: Question:

Is this accurate? Just curious because it seems odd. If that's the case, why would you ever need to include a field? I might be missing something. This also doesn't match `$project`, but I am beginning to think that is expected.

source/fundamentals/crud/read-operations/project.txt
```diff
@@ -57,149 +74,132 @@ method in the options of the following read operation methods:
...
-
-.. important::
-
-    You can exclude the ``_id`` field even if you specified to include
-    certain fields. By default, the driver includes the ``_id`` field.
-    You must explicitly exclude the ``_id`` field if you do not want it
-    returned.
+To include a field, pass the field you want to include with a ``1`` to
+the ``SetProjection()`` method. The driver excludes all fields that are
+not explicitly included in the projection document.
```
DBirtolo-mdb, line 123: Question:

Now I am very confused. I feel like something is missing from this description and above. Because we now both say that we include all fields not explicitly excluded and that we exclude all fields not explicitly included. These cannot both be true. 

Does this need something additional like "...if you specify any fields to include"?

DBirtolo-mdb (changes requested):
Some questions about the `SetProjection` method that might not be anything that requires an update. Also seems like things that existed before your updates, but I still had to ask.

source/fundamentals/crud/read-operations/project.txt
```diff
@@ -21,27 +21,44 @@ document from read operations.
...
+the ``title``, ``course_id``, and ``enrollment`` fields in each document.
 
 Projection
 ----------
 
-A projection specifies which fields to return in matched documents. It
-contains field names followed by a ``1`` (to include) or ``0`` (to
-exclude). Projections can only include or exclude fields.
+A projection specifies which fields to return in matched documents. The
+projection document contains field names with a ``1`` to include the
```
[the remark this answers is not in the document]
rustagir: Yes, I think that line existed before I made these edits. I agree this line should be edited.

source/fundamentals/crud/read-operations/project.txt
rustagir: I think the purpose is for documents with many fields, it is faster to either include or exclude just a few fields instead of the counterpart. For example, if a document had 30 fields and you only wanted to include two fields in a projection, you could specify an include, instead of excluding 28 fields. For the example data I used in this page (only 3 fields), I can see why it seems unclear.

DBirtolo-mdb (approved):
LGTM! Thanks for clearing that up!

PR #225, opened 2023-01-17
DOCSP-26229: improve-runcommand

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-26229
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-26229-improve-runcmd/fundamentals/crud/run-command/

See #178 for comments that this PR addresses.

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/crud/run-command.txt
+++ b/source/fundamentals/crud/run-command.txt
@@ -93,10 +93,10 @@ following fields:
      - Indicates the logical time of the operation. MongoDB uses the
        logical time to order operations.
        
-     .. seealso::
-
-        To learn more about logical time, see our :website:`blog post about
-        the Global Logical Clock </blog/post/transactions-background-part-4-the-global-logical-clock>`.
+       .. seealso::
+  
+          To learn more about logical time, see our :website:`blog post about
+          the Global Logical Clock </blog/post/transactions-background-part-4-the-global-logical-clock>`.
 
    * - ``$clusterTime``
      - Provides a document that returns the signed cluster time. Cluster time is a
--- a/source/fundamentals/crud/run-command.txt
+++ b/source/fundamentals/crud/run-command.txt
@@ -15,16 +15,19 @@ Run a Command
 Overview
 --------
 
-In this guide, you can learn how to run a **database command** with the
+In this guide, you can learn how to run a database command with the
 {+driver-short+}. You can use database commands to perform a variety of
 administrative and diagnostic tasks, such as fetching server statistics,
 initializing a replica set, or running an aggregation pipeline.
 
-Compose a Command
+Execute a Command
 -----------------
 
-The {+driver-short+} provides the following methods to run database
-commands: 
+To run a database command, you must specify the command and any relevant
+parameters in a command document, then pass the command document to a
+wrapper method. The command document must be an order-preserving type
+such as ``bson.D``. The {+driver-short+} provides the following methods
+to run database commands:
 
 - ``RunCommand()``, which returns the command response as a
   ``SingleResult`` type. You can use this method with any database command.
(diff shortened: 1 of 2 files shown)


source/fundamentals/crud/run-command.txt
```diff
@@ -15,46 +15,50 @@ Run a Command
...
+
+For a full list of database commands and corresponding parameters, see
+the :ref:`Additional Information section <addl-info-runcommand>`.
 
 .. note:: Read Preference
 
    ``RunCommand()`` and ``RunCommandCursor()`` do not obey the read
-   preference set for the database. You can set a read preference for a
-   database command by passing a ``RunCmdOptions`` type to either method:
+   preference you set on your ``Database`` object. You can set a read preference for a
```
mongoKart, line 55: Issue: If someone else has set the read preference in a different way, does ``RunCommand()`` obey that? As written, it sounds like it will ignore just the one that I set.

Suggestion: Clarify scope of what ``RunCommand()`` will respect.

source/fundamentals/crud/run-command.txt
```diff
@@ -63,7 +67,10 @@ Information <addl-info-runcommand>`.
 Response
 --------
 
-Each method returns a document or a cursor containing documents with the
+Each method returns a result or a cursor that contains the response
```
mongoKart, line 70: Issue: What format is the result?

source/fundamentals/crud/run-command.txt
```diff
@@ -63,7 +67,10 @@ Information <addl-info-runcommand>`.
...
+from the database after you run a command. Each database command
```
mongoKart, line 71: Issue: "after you run a command" is a little vague

Suggestion: I know we try to avoid passive, but I think it makes sense here: "...from the database after the command has completed." (or something similar)

source/fundamentals/crud/run-command.txt
```diff
@@ -63,7 +67,10 @@ Information <addl-info-runcommand>`.
...
+requests different information, so the response content
```
mongoKart, line 72: I: Is every command just a query (requesting info from the DB) or are there other types too?

S: If the commands do very different things, maybe change to "Each database command performs a different function" or similar

source/fundamentals/crud/run-command.txt
```diff
@@ -114,32 +125,46 @@ run the ``count`` command on the ``flowers`` collection of the
 Output
 ~~~~~~
 
-In the output, you should see the number of documents in
-the collection that match the query as the value of the ``n`` field, as
-well as the command execution information:
+In the output, you should see fields explaining the
+execution of the count operation, such as the winning plan and rejected
+plans. The output also contains the explain command execution information:
```
mongoKart, line 130: Issue: I could see this being a little hard to parse, especially for someone not familiar with explain commands. For example, do we expect the reader to know what a "winning plan" is in this context?

I also initially had trouble parsing the last sentence. Maybe change to "The output also contains information about the execution of the ``explain`` command."

mongoKart (changes requested):
Just a couple small things

source/fundamentals/crud/run-command.txt
rustagir: What this means is that when you instantiate a database object, you can set a read preference. But, the RunCommand method(s) will not obey that read preference.
I think the change I made improves clarity. LMK

source/fundamentals/crud/run-command.txt
rustagir: [Some commands](https://www.mongodb.com/docs/manual/reference/command/nav-crud/) do make changes to the database/collection. I will change the language accordingly.

source/fundamentals/crud/run-command.txt
rustagir: I added a clause explaining what a winning plan is.

mongoKart (approved):
LGTM

PR #226, opened 2023-01-17
DOCSP-13871: FAQ page

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-13871
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-13871-FAQ/faq/

## Self-Review Checklist

### DIFF
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -55,7 +55,7 @@ the ``FindOneAndUpdate()`` operation:
        panic(err)
    }
 
-The following code shows how to correctly initialize the ``sort``
+The following code shows how to correctly initialize the ``sortOrder``
 variable as a ``bson.D`` type so that the driver can convert it to BSON:
 
 .. code-block:: go
@@ -106,8 +106,8 @@ Why Did the Driver Throw an "Authentication failed" Error?
 If you attempt to connect to a MongoDB deployment using incorrect
 username and password credentials or specifying the wrong authentication
 mechanism, you receive an "Authentication failed" error message. For
-example, the following error message is returned when ``SCRAM-SHA-256``
-authentication fails:
+example, the following error message is returned when you pass invalid
+credentials when authenticating with ``SCRAM-SHA-256``:
 
 .. code-block:: none
    :copyable: false
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -6,12 +6,118 @@ FAQ
 
 .. default-domain:: mongodb
 
-Frequently Asked Questions
-
 .. contents:: On this page
    :local:
    :backlinks: none
-   :depth: 1
+   :depth: 2
    :class: singlecol
 
-Lorem ipsum.
+.. tip::
+
+   This page contains frequently asked questions and their corresponding
+   answers. If you can't find an answer to your problem on this page,
+   see the :ref:`golang-issues-and-help` page for next steps and more
+   resources.
+   
+How Can I Fix the "``WriteXXX()`` can only write while positioned on an Element or Value but is positioned on a TopLevel" Error?
+--------------------------------------------------------------------------------------------------------------------------------
+
+The ``bson.Marshal()`` method requires a parameter that can be decoded
+into a BSON Document, such as a ``primitive.D`` type. This error occurs
+when the input to ``bson.Marshal()`` is something *other* than a BSON
+Document.
+
+The following list includes some example situations that surface this error:
+
+- You pass a string to ``bson.Marshal()``, causing a ``WriteString`` error.
+- You pass a ``null`` to ``bson.Marshal()``, causing a ``WriteNull`` error.
+- You pass a string to ``bson.Marshal()``, causing a ``WriteInt32`` error.
+
+Many CRUD operations use the ``bson.Marshal()`` method internally,
+so you may encounter this error even if you are not directly attempting
+to encode data.
+
+The following code produces this error because of how the ``sort``
+variable is initialized:
+
+.. code-block:: go
+
+   var sort bson.D // this is null and will result in a WriteNull error
+   opts := options.FindOneAndUpdate().SetSort(sort)
+
+   update := bson.D{{"$inc", bson.D{{"counter", 1}}}}
+   result := coll.FindOneAndUpdate(ctx, bson.D{}, update)
+   if err := result.Err(); err != nil {
+   	log.Fatalf("error getting single result: %v", err)
+   }
+
+The following code shows how to correctly initialize the ``sort``
+variable as a ``bson.D`` type:
+
+.. code-block:: go
+
+   sort := bson.D{}
+
+How Do I Convert a BSON Document to JSON?
+-----------------------------------------
+
+The driver provides a variety of marshaller methods that can be used to
+encode a BSON document as JSON, such as the ``MarshalExtJSON()`` method:
+
+.. io-code-block::
+   :copyable: true
+
+   .. input::
+      :language: go
+      :emphasize-lines: 3
+
+      jsonBytes, err := bson.MarshalExtJSON(doc, true, false)
+      if err != nil {
+      	log.Fatalf("error encoding: %v", err)
+      }
+      
+      m := make(map[string]interface{})
+      if err := json.Unmarshal(jsonBytes, &m); err != nil {
+      	log.Fatalf("error decoding: %v", err)
+      }
+      fmt.Printf("json: %v\n", m)
+
+   .. output::
+      :language: none
+      :visible: false
+
+      json: map[hello:world]
+
+To learn more about encoding and decoding, see the
+:ref:`golang-bson` guide.
+
+Why Did My Authentication Fail?
+-------------------------------
+
+When connecting to a MongoDB deployment that requires password
+authentication, you may get an error message if any credentials or the
+authentication configuration are incorrect. For example, the following
+error message is returned when ``SCRAM-SHA-256`` authentication fails:
+
+.. code-block:: none
+
+   connection() error occurred during connection handshake: auth error:
+   sasl conversation error: unable to authenticate using mechanism
+   "SCRAM-SHA-256": (AuthenticationFailed) Authentication failed.
+
+The error message intentionally omits the exact authentication failure
+reason, but common causes for authentication failure include the
+following:
+
+- Incorrect password
+- Incorrect username
+- Incorrect specification of authentication method in the
+  ``AuthSource`` property
+
+To avoid this error, ensure that the username and password in your
+connection string or ``SetAuth()`` call are correct. In most cases,
+you do not need to specify ``AuthSource`` unless you are using a
+non-default authentication database.
+
+To learn more about authentication, see the
+:ref:`golang-authentication-mechanisms` guide.
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -20,18 +20,20 @@ This page contains frequently asked questions and their corresponding answers.
    see the :ref:`golang-issues-and-help` page for next steps and more
    resources.
    
-How Can I Fix the "WriteXXX can only write while positioned on a Element or Value but is positioned on a TopLevel" Error?
--------------------------------------------------------------------------------------------------------------------------
+How Can I Fix the "WriteNull can only write while positioned on a Element or Value but is positioned on a TopLevel" Error?
+--------------------------------------------------------------------------------------------------------------------------
 
 The ``bson.Marshal()`` method requires a parameter that can be decoded
 into a BSON document, such as the ``bson.D`` type. This error occurs
-when the input to ``bson.Marshal()`` is something *other* than a BSON
-document.
+when you pass something *other* than a BSON document to
+``bson.Marshal()``.
 
-Situations in which this error can occur include the following:
+The ``WriteNull`` error occurs when you pass a ``null`` to
+``bson.Marshal()``. Situations in which a similar error can occur
+include the following:
 
 - You pass a string to ``bson.Marshal()``, causing a ``WriteString`` error.
-- You pass a ``null`` to ``bson.Marshal()``, causing a ``WriteNull`` error.
+- You pass a boolean to ``bson.Marshal()``, causing a ``WriteBoolean`` error.
 - You pass an integer to ``bson.Marshal()``, causing a ``WriteInt32`` error.
 
 You may encounter this error when you perform a CRUD operation that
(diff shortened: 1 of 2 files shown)


source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
+
+.. code-block:: none
+   :copyable: false
+
+   connection() error occurred during connection handshake: auth error:
+   sasl conversation error: unable to authenticate using mechanism
+   "SCRAM-SHA-256": (AuthenticationFailed) Authentication failed.
+
+The error message intentionally omits the exact authentication failure
+reason for security purposes, but common causes for authentication failure include the
```
ccho-mongodb, line 112: Suggestion:
The error message can't have an intention, although the author of the code that generates it can. Avoid personifying inanimate objects when possible. E.g.

This error message does not contain the detailed authentication failure reason for security reasons, but indicates the issue occurred for one of the following reasons:

source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
+      if err := json.Unmarshal(jsonBytes, &m); err != nil {
+      	log.Fatalf("error decoding: %v", err)
+      }
+      fmt.Printf("json: %v\n", m)
+
+   .. output::
+      :language: none
+      :visible: false
+
+      json: map[hello:world]
```
ccho-mongodb, line 91: Question: The description of this code block mentions encoding a BSON document to JSON, but isn't this a Go map? I would have expected JSON to look like:
```json
 { "hello": "world" }
```

source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
+reason for security purposes, but common causes for authentication failure include the
+following:
+
+- Incorrect password
+- Incorrect username
+- Incorrect specification of authentication method in the
+  ``AuthSource`` property
+
+To avoid this error, ensure that the username and password in your
+connection string or ``SetAuth()`` call are correct. In most cases,
```
ccho-mongodb, line 121: Issue: 

1) The "or SetAuth() call are correct" doesn't seem parallel to "username and password in your connection string", but should be.
2) Avoid recommending putting your credentials in plaintext in the connection string because that is a security risk. If the driver supports providing credentials in environment variables, suggest that instead.

Suggestion:
Start with the general case (credentials are correct) and then describe two specific ways to achieve the general case. E.g. 
To avoid this error, make sure the credentials you provided are correct. You can set your authentication credentials in environment variables or by passing them to the SetAuth() method.

source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
+   sasl conversation error: unable to authenticate using mechanism
+   "SCRAM-SHA-256": (AuthenticationFailed) Authentication failed.
+
+The error message intentionally omits the exact authentication failure
+reason for security purposes, but common causes for authentication failure include the
+following:
+
+- Incorrect password
+- Incorrect username
+- Incorrect specification of authentication method in the
```
ccho-mongodb, line 117: Issue: the Go driver refers to this as authentication mechanism instead of method.

Suggestion:
Slightly less verbose version:

"Incorrect authentication mechanism specified in the AuthSource property"

ccho-mongodb (changes requested):
Identified a few issues and made suggestions.

If you haven't already done so, could you make sure of the following:
- The code examples run with the latest driver (ensure the method signatures haven't changed and accomplish the stated goals)
- The error messages are still the same in the latest driver
- The code examples follow the coding practices we agreed upon with DBX for use in our driver documentation

Thanks!

source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
-
 .. contents:: On this page
    :local:
    :backlinks: none
-   :depth: 1
+   :depth: 2
    :class: singlecol
 
-Lorem ipsum.
+.. tip::
```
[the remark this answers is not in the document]
rustagir: moved first sentence to normal text, left link to I&H in tip

source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
+
+   This page contains frequently asked questions and their corresponding
+   answers. If you can't find an answer to your problem on this page,
+   see the :ref:`golang-issues-and-help` page for next steps and more
+   resources.
+   
+How Can I Fix the "WriteXXX() can only write while positioned on an Element or Value but is positioned on a TopLevel" Error?
```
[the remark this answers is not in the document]
rustagir: This is the error message produced by the code example:
`WriteNull can only write while positioned on a Element or Value but is positioned on a TopLevel`. 
So, WriteXXX is a placeholder as the error message depends on the type passed to `bson.Marshal()`.

source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
+The ``bson.Marshal()`` method requires a parameter that can be decoded
+into a BSON Document, such as a ``bson.D`` type. This error occurs
+when the input to ``bson.Marshal()`` is something *other* than a BSON
+Document.
+
+The following list includes some situations that surface this error:
+
+- You pass a string to ``bson.Marshal()``, causing a ``WriteString`` error.
+- You pass a ``null`` to ``bson.Marshal()``, causing a ``WriteNull`` error.
+- You pass a string to ``bson.Marshal()``, causing a ``WriteInt32`` error.
```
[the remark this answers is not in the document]
rustagir: oops, that is a mistake. changed string to integer in the third item.

source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
+      }
+      fmt.Printf("json: %v\n", m)
+
+   .. output::
+      :language: none
+      :visible: false
+
+      json: map[hello:world]
+
+To learn more about encoding and decoding with BSON, see the
```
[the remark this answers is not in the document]
rustagir: changed wording to clarify.

source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
+
+The following list includes some situations that surface this error:
+
+- You pass a string to ``bson.Marshal()``, causing a ``WriteString`` error.
+- You pass a ``null`` to ``bson.Marshal()``, causing a ``WriteNull`` error.
+- You pass a string to ``bson.Marshal()``, causing a ``WriteInt32`` error.
+
+Many CRUD operations use the ``bson.Marshal()`` method internally,
+so you may encounter this error even if you are not directly attempting
+to encode data.
```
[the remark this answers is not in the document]
rustagir: I'm not sure I completely understand what you mean by "seems to imply that their usage was directly attempting to encode data". I think you overall meant that it will be easier for users to understand the two main situations when this error arises: when they directly call `bson.Marshal()` and when they perform a CRUD operation that internally uses this method. Changed wording to reflect this.

source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
+
+The following code produces this error because of how the ``sort``
+variable is initialized:
```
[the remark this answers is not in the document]
rustagir: I reworded to explain what exactly surfaces the error

source/faq.txt
```diff
@@ -6,12 +6,121 @@ FAQ
...
+
+To learn more about encoding and decoding with BSON, see the
+:ref:`golang-bson` guide.
+
+Why Did My Authentication Fail?
+-------------------------------
+
+When connecting to a MongoDB deployment that requires password
+authentication, you may get an error message if any credentials or the
+authentication configuration are incorrect. For example, the following
```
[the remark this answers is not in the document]
rustagir: It is not only incorrect credentials that can cause an error-- it can also be specifying the authentication method incorrectly. I changed the wording to reflect that an error **will** be surfaced if either of these situations are applicable.

source/faq.txt
rustagir: the error message does not indicate why the issue occurred. changed wording and broke up sentence.

source/faq.txt
rustagir: good point! changed the explanation and the example and output

source/faq.txt
```diff
@@ -6,12 +6,126 @@ FAQ
...
+- You pass a ``null`` to ``bson.Marshal()``, causing a ``WriteNull`` error.
+- You pass an integer to ``bson.Marshal()``, causing a ``WriteInt32`` error.
+
+You may encounter this error when you perform a CRUD operation that
+internally uses the ``bson.Marshal()`` method or when you call
+``bson.Marshal()`` directly to encode data.
+
+The following code produces a ``WriteNull`` error because the driver
+cannot encode the ``null`` valued ``sortOrder`` variable to BSON during
+the find-and-update operation:
```
ccho-mongodb, line 43: Suggestion: 
Unless "find-and-update" is a common term for the Go driver (I haven't seen it used in other drivers), it could be more clear to refer to it by the method name instead:

```suggestion
the FindOneAndUpdate() operation:
```

source/faq.txt
```diff
@@ -6,12 +6,126 @@ FAQ
...
+   sasl conversation error: unable to authenticate using mechanism
+   "SCRAM-SHA-256": (AuthenticationFailed) Authentication failed.
+
+This error message does not contain the detailed authentication failure
+reason for security reasons. Common causes for authentication failure include the
+following:
+
+- Incorrect password
+- Incorrect username
+- Incorrect authentication mechanism specified in the ``AuthSource`` property
```
ccho-mongodb, line 122: Question:
Is it only in the AuthSource property or could it be in the connection string?

source/faq.txt
```diff
@@ -6,12 +6,126 @@ FAQ
...
+
+To avoid this error, ensure that your credentials and authentication
+mechanism are correct. You can store your authentication credentials in
+environment variables or you can pass them to the ``SetAuth()`` method.
+In most cases, you do not need to set the ``AuthSource`` property unless
+you are using a non-default authentication database.
```
ccho-mongodb, line 128: Suggestion:
This sentence seems to provide information that isn't related to the main topic of why someone received this error message and how to resolve it.

If it seems essential, I think "In most cases" makes it unclear as to what other reasons you would need to set the AuthSource property other than what's listed, and would be helpful to list the other cases.

source/faq.txt
```diff
@@ -6,12 +6,126 @@ FAQ
...
+   sort := bson.D{}
+
+How Do I Convert a BSON Document to JSON?
+-----------------------------------------
+
+The driver provides a variety of marshaller methods that can be used to
+convert a BSON document to JSON, such as the ``MarshalExtJSON()``
+method. To view a readable form of the JSON encoding, you must use
+an unmarshaller method or string type-casting to parse the JSON byte
+format.
```
ccho-mongodb, line 70: 👍

source/faq.txt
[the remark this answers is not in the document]
ccho-mongodb: Issue:
I think we should avoid "XXX" as a placeholder since it's not a widely used that way and doesn't make it more searchable per my comment.

Suggestion:
Choose one of the "Write...()" operations such as "WriteNull" and use that as the title. Then, in the answer to this question, mention some or all of the "Write...()" operations to which this could apply.

source/faq.txt
```diff
@@ -6,12 +6,126 @@ FAQ
...
+The following code produces a ``WriteNull`` error because the driver
+cannot encode the ``null`` valued ``sortOrder`` variable to BSON during
+the find-and-update operation:
+
+.. code-block:: go
+
+   var sortOrder bson.D
+   opts := options.FindOneAndUpdate().SetSort(sortOrder)
+   
+   update := bson.D{{"$inc", bson.D{{"counter", 1}}}}
```
ccho-mongodb, line 50: Similar comment to https://github.com/mongodb/docs-golang/pull/226#discussion_r1072938457

source/faq.txt
```diff
@@ -6,12 +6,126 @@ FAQ
...
+   if err := result.Err(); err != nil {
+       panic(err)
+   }
+
+The following code shows how to correctly initialize the ``sort``
+variable as a ``bson.D`` type so that the driver can convert it to BSON:
+
+.. code-block:: go
+
+   sort := bson.D{}
```
ccho-mongodb, line 61: Similar comment to https://github.com/mongodb/docs-golang/pull/226#discussion_r1072938457

source/faq.txt
```diff
@@ -6,12 +6,126 @@ FAQ
...
+      
+      fmt.Println(string(jsonBytes))
+
+   .. output::
+      :language: none
+      :visible: false
+
+      {"hello":"world"}
+
+To learn more about conversions between BSON and Go values, see the
```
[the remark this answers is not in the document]
rustagir: I was using the phrasing from the Work with BSON page, but I think that that should be changed there too then. LMK if you think that I should change that in this PR or if it is out of scope.

source/faq.txt
rustagir: authSource (lowercase a) is the connection string option, and AuthSource is the property of the Credential struct. I changed the wording to reflect this

source/faq.txt
rustagir: I think it's best to just omit this sentence, I think this info is given on both authentication fundamentals pages anyways.

source/faq.txt
rustagir: changed to `updateDocument`

source/faq.txt
rustagir: thx forgot to change it here too!

source/faq.txt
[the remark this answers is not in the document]
rustagir: changed answer to address specific error but indicated the causes of similar errors (only differ by type)

source/faq.txt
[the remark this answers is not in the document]
rustagir: this makes sense! I didn't understand that you meant to move the conditions to the first clause of the sentence

source/faq.txt
```diff
@@ -6,12 +6,128 @@ FAQ
...
+:ref:`golang-bson` guide.
+
+Why Did the Driver Throw an "Authentication failed" Error?
+----------------------------------------------------------
+
+If you attempt to connect to a MongoDB deployment using incorrect
+username and password credentials or specifying the wrong authentication
+mechanism, you receive an "Authentication failed" error message. For
+example, the following error message is returned when ``SCRAM-SHA-256``
+authentication fails:
```
ccho-mongodb, line 110: Suggestion: 

Connect the previous sentence about what potential causes are possible with one specific stated cause:

```suggestion
example, the following error message is returned when you
pass invalid credentials when authenticating with ``SCRAM-SHA-256``:
```

source/faq.txt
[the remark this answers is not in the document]
ccho-mongodb: After looking at the Work with BSON page, there are a couple instances in which "type[s]" would be preferable. I think a ticket would probably be better since that task doesn't seem central to publishing the FAQ.

source/faq.txt
```diff
@@ -6,12 +6,128 @@ FAQ
...
+   var sortOrder bson.D
+   opts := options.FindOneAndUpdate().SetSort(sortOrder)
+   
+   updateDocument := bson.D{{"$inc", bson.D{{"counter", 1}}}}
+   result := coll.FindOneAndUpdate(context.TODO(), bson.D{}, updateDocument, opts)
+   if err := result.Err(); err != nil {
+       panic(err)
+   }
+
+The following code shows how to correctly initialize the ``sort``
```
ccho-mongodb, line 58:
```suggestion
The following code shows how to correctly initialize the ``sortOrder``
```

ccho-mongodb (approved):
LGTM, just a few suggestions.

PR #238, opened 2023-02-17
DOCSP-27920: atlas connection snippet w stableapi

rustagir:
# Pull Request Info

[PR Reviewing Guidelines](https://github.com/mongodb/docs-golang/blob/master/REVIEWING.md)

JIRA - https://jira.mongodb.org/browse/DOCSP-27920
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-27920-quickstart-stable-api/fundamentals/connection/

**Note: I replaced the connection string with a placeholder to align with guidance from [this document](https://docs.google.com/document/d/1m-mEe8qH_rIAAiekC8Ehr4qjpSY-BOOapJAyizsIoHM/edit#)**

## Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST?
- [x] Did you run a spell-check?
- [ ] Did you run a grammar-check?
- [x] Are all the links working?

PR #239, opened 2023-02-22
Respond to CC feedback on #238

rustagir:
https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/CC-feedback-27920/fundamentals/connection/#connection-example

#238

### DIFF
--- a/source/fundamentals/connection.txt
+++ b/source/fundamentals/connection.txt
@@ -19,10 +19,13 @@ deployment using the Go Driver.
 Connection URI
 --------------
 
-The **connection URI** provides a set of instructions that the driver uses to
-connect to a MongoDB deployment. It instructs the driver on how it should
-connect to MongoDB and how it should behave while connected. The following
-example explains each part of a sample connection URI:
+The **connection URI**, also known as a connection string, tells the
+driver how to connect to MongoDB and how to behave while connected.
+
+Parts of a Connection URI
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following example explains each part of a sample connection URI:
 
 .. figure:: /includes/figures/connection_uri_parts.png
    :alt: Each part of the connection string
@@ -55,25 +58,30 @@ options. In the example, we set two connection options:
 ``maxPoolSize=20`` and ``w=majority``. To learn more about connection
 options, read the :ref:`golang-connection-options` section of this guide.
 
-Create a Client to Connect to MongoDB
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Connection Example
+~~~~~~~~~~~~~~~~~~
 
-You need to create a client to connect to MongoDB, manage your
-connection, and run database commands. You can create a client by
-passing options to the ``Connect()`` method. These options
-include your connection string and any other client options.
+To connect to MongoDB, you need to create a client. A client manages
+your connections and runs database commands.
 
-To learn more about creating a client in the {+driver-short+}, see the
-API documentation for `Client <{+api+}/mongo#Client>`__ and `Connect()
-<{+api+}/mongo#Connect>`__.
+You can create a client by calling the ``Connect()`` method. You can
+specify your connection string and other client options by using the
+``options`` package. You can specify your connection URI by passing it
+to the ``ApplyURI()`` method. You can set any other options by using
+methods from the options package.
 
-In the options, you can set a Stable API server version to avoid
+To learn more about connection options, see the
+:ref:`Connection Options section <golang-connection-options>`. To learn
+more about creating a client, see the API documentation for `Client
+<{+api+}/mongo#Client>`__ and `Connect() <{+api+}/mongo#Connect>`__.
+
+You can set the {+stable-api+} version as an option to avoid
 breaking changes when you upgrade to a new server version. To 
-learn more about the Stable API feature, see the :ref:`Stable API page
+learn more about the {+stable-api+} feature, see the :ref:`{+stable-api+} page
 <golang-stable-api>`.
 
 The following code shows how you can create a client that uses your
-connection string and a Stable API version, connect to MongoDB, and
+connection string and the {+stable-api+} version, connect to MongoDB, and
 verify that the connection is successful:
 
 .. literalinclude:: /includes/fundamentals/code-snippets/srv.go
--- a/source/fundamentals/connection.txt
+++ b/source/fundamentals/connection.txt
@@ -19,7 +19,7 @@ deployment using the Go Driver.
 Connection URI
 --------------
 
-The **connection URI**, also known as a connection string, tells the
+A **connection URI**, also known as a connection string, tells the
 driver how to connect to MongoDB and how to behave while connected.
 
 Parts of a Connection URI
@@ -64,11 +64,11 @@ Connection Example
 To connect to MongoDB, you need to create a client. A client manages
 your connections and runs database commands.
 
-You can create a client by calling the ``Connect()`` method. You can
-specify your connection string and other client options by using the
-``options`` package. You can specify your connection URI by passing it
-to the ``ApplyURI()`` method. You can set any other options by using
-methods from the options package.
+You can create a client that uses your connection string and other
+client options by passing a ``ClientOptions`` object to the ``Connect()``
+method. To specify your connection URI, pass it to the ``ApplyURI()``
+method, creating a ``ClientOptions`` instance. To set any other
+options, call the relevant helper method from the ``options`` package.
 
 To learn more about connection options, see the
 :ref:`Connection Options section <golang-connection-options>`. To learn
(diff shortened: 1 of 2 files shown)


source/fundamentals/connection.txt
```diff
@@ -55,10 +55,15 @@ options. In the example, we set two connection options:
 ``maxPoolSize=20`` and ``w=majority``. To learn more about connection
 options, read the :ref:`golang-connection-options` section of this guide.
 
-The following code shows how you can use a connection string in a
-client to connect to MongoDB. In this example, the client uses 
-the :ref:`Stable API <golang-stable-api>` feature, though
-you can connect to MongoDB without this feature.
+To connect to MongoDB, create a client by passing your connection string
```
ccho-mongodb, line 58: Suggestion:
On a second read, I think the term "client" needs an introduction and this seems to be a good place to transition. I think one of the better solutions is to separate the explanation of how the Connection URI relates to the client into a sub-section since this one is about the Connection URI.

E.g.


```
MongoDB Client
~~~~~~~~~~~~~~

You need to create a MongoDB client to manage your connection 
and run database commands. To create a client, call the ... method 
and pass it your connection string and any options...

< perhaps link to some docs about the client and options >

In the options, you can set the Stable API version. The Stable API 
option ensures that your application runs commands that are
compatible with future server versions... Learn more about...
```

source/fundamentals/connection.txt
```diff
@@ -55,10 +55,15 @@ options. In the example, we set two connection options:
...
+and any options. In the options, you can set a Stable API server version
+to avoid breaking changes when you upgrade to a new server version. To
+learn more about the Stable API feature, see the :ref:`Stable API page
+<golang-stable-api>`.
+
+The following code shows how you can instantiate a client that uses your
```
ccho-mongodb, line 64: Suggestion: 

I think it could be good to use consistent terminology whether "create" or "instantiate".
```suggestion
The following code shows how you can create a client that uses your
```

ccho-mongodb (changes requested):
Looks pretty good -- from looking at the page in this perspective, I had a few new thoughts to help organize this info.

A couple additional items outside of the text included int his PR:

Issue 1:
I think there's a disconnect that's never explained about the equivalence of "serverAPI" and "Stable API".

Suggestion 1:

I think this adding a code comment to identify the code related to Stable API could be helpful. The comment needs to be standalone since it will be in the Atlas UI which I don't think provides any explanation.


Issue 2:

Similar to issue 1, I think the comment "Ping the primary" should be more self-explanatory because it is going to be used int he Atlas UI which I don't think provides any explanation.

Suggestion 2:
Perhaps something like "Send a ping to the primary node of your cluster of to check whether you successfully connected". Or alternatively, update the comment to something like "Send a ping to confirm the connection" and update the Println message to explain it. "Successfully connected and received a ping acknowledgement from the MongoDB cluster."

source/fundamentals/connection.txt
```diff
@@ -55,10 +55,26 @@ options. In the example, we set two connection options:
 ``maxPoolSize=20`` and ``w=majority``. To learn more about connection
 options, read the :ref:`golang-connection-options` section of this guide.
 
-The following code shows how you can use a connection string in a
-client to connect to MongoDB. In this example, the client uses 
-the :ref:`Stable API <golang-stable-api>` feature, though
-you can connect to MongoDB without this feature.
+Create a Client to Connect to MongoDB
```
ccho-mongodb, line 58: Suggestion:
Maybe the information on this page could be organized as follows:

Connection URI
- Compose a Connection String
- Create a Client to Connect to MongoDB

source/fundamentals/connection.txt
```diff
@@ -55,10 +55,26 @@ options. In the example, we set two connection options:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You need to create a client to connect to MongoDB, manage your
+connection, and run database commands. You can create a client by
+passing options to the ``Connect()`` method. These options
+include your connection string and any other client options.
```
ccho-mongodb, line 64: Issue:

I think there's ambiguity in the current presentation of "options" that might only be specific to the Golang driver.

In the driver, there's an "option" package from which you access the Client and include your connection string via ApplyURI(). However, the general term "option" describes something that you can omit (and doesn't have a default value) and have it function properly. I think the Client() contains a default value (mongodb://localhost:27017) when you just instantiate with "mongo.NewClient()". This allows you to connect locally, but it would be good to identify as required in all cases except when you're connecting locally to the default port, especially since the prior content describes the connection string in detail. 

Suggestion:

You can create a client by calling the mongo.NewClient() method. You can specify your connection string and other connection options using the options package. To specify your connection string, call the method ApplyURI() on the options package. To specify <other options such as... > call < other methods >. Pass your options to the Connect() method to connect to the MongoDB deployment that you specified in your connection string.

source/fundamentals/connection.txt
```diff
@@ -55,10 +55,26 @@ options. In the example, we set two connection options:
...
+
+To learn more about creating a client in the {+driver-short+}, see the
+API documentation for `Client <{+api+}/mongo#Client>`__ and `Connect()
+<{+api+}/mongo#Connect>`__.
+
+In the options, you can set a Stable API server version to avoid
```
ccho-mongodb, line 70: Suggestion:

Avoid the options package ambiguity mentioned in the other comment.
```suggestion
You can set a Stable API server version option to avoid
```

source/includes/fundamentals/code-snippets/srv.go
```diff
@@ -29,10 +30,10 @@ func main() {
 		}
 	}()
 
-	// Ping the primary
+	// Send a ping to confirm successful connection
```
ccho-mongodb, line 33: Suggestion:

I think this needs an article since "successful connection" is an object rather than the name of a status.

```suggestion
	// Send a ping to confirm a successful connection
```

source/includes/fundamentals/code-snippets/srv.go
```diff
@@ -29,10 +30,10 @@ func main() {
...
 	if err := client.Ping(context.TODO(), readpref.Primary()); err != nil {
 		panic(err)
 	}
 
-	fmt.Println("Successfully connected and pinged.")
+	fmt.Println("Pinged the primary node of the cluster. You are connected to MongoDB!")
```
ccho-mongodb, line 38: Suggestion:

I wonder if it might be confusing to mention that "you are connected" and then immediately disconnect after that. What do you think about "You successfully connected to MongoDB!" to replace the second sentence?

source/includes/fundamentals/code-snippets/srv.go
```diff
@@ -14,6 +14,7 @@ const uri = "<connection string>"
 
 func main() {
 
+	// Use the ServerAPI() method to set Stable API version 1
```
ccho-mongodb, line 17: Suggestion:

I like the brevity of the comment, but think this fragment still needs an article and preposition to identify what's being set and what it's set to.

```suggestion
	// Use the ServerAPI() method to set the Stable API to version 1
```

ccho-mongodb (changes requested):
Very good improvements! I think there's an issue with the newly-introduced content -- let me know if you want to pair on this.

source/fundamentals/connection.txt
```diff
@@ -55,10 +58,31 @@ options. In the example, we set two connection options:
...
+~~~~~~~~~~~~~~~~~~
+
+To connect to MongoDB, you need to create a client. A client manages
+your connections and runs database commands.
+
+You can create a client by calling the ``Connect()`` method. You can
+specify your connection string and other client options by using the
+``options`` package. You can specify your connection URI by passing it
+to the ``ApplyURI()`` method. You can set any other options by using
+methods from the options package.
```
ccho-mongodb, line 71: Suggestion:

The last two sentences seem to have some redundancies with the second sentence, so I think it can be made more concise. Also, keep consistent with the formatting for the "options package" so that it's less likely to get confused with "[client] options".

E.g.

You can pass your connection string and other client options to the Connect() method using methods in the options package. To specify your connection URI, pass it to the ApplyURI() method. To specify other options, call the appropriate helper method from the options package. (note that you should apply monospace where appropriate)

source/includes/fundamentals/code-snippets/srv.go
```diff
@@ -14,6 +14,7 @@ const uri = "<connection string>"
 
 func main() {
 
+	// Use the ServerAPI() method to set the Stable API version to 1
```
ccho-mongodb, line 17: Suggestion:

Sorry for missing this -- shouldn't it be the SetServerAPIOptions method that actually sets the Stable API version? I think options.ServerAPI() returns a ServerAPIOptions pointer.

```suggestion
	// Use the SetServerAPIOptions() method to set the Stable API version to 1
```

ccho-mongodb (approved):
LGTM, a couple suggestions

PR #277, opened 2023-07-07
Docsp 29902 revise quick reference examples

nickldp:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-29902>
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/golang/docsworker-xlarge/DOCSP-29902-Revise-quick-reference-examples/quick-reference/)

## Self-Review Checklist

### DIFF
--- a/source/quick-reference.txt
+++ b/source/quick-reference.txt
@@ -131,8 +131,8 @@ their related reference and API documentation.
 
              result, err := coll.UpdateMany(
                  context.TODO(), 
-                 bson.D{{"age", bson.D{{"$lt", 10}}}}, 
-                 bson.D{{"$set", bson.D{{"age", 1}}}}
+                 bson.D{{"age", bson.D{{"$gte", 58}}}}, 
+                 bson.D{{"$set", bson.D{{"description", "Senior"}}}}
              )
              fmt.Printf("The number of modified documents: %d\n", result.ModifiedCount)
 
@@ -441,7 +441,7 @@ their related reference and API documentation.
              :language: go
              :visible: false 
 
-             [{firstName Lester} ]
+             [{firstName Lester}]
              [{firstName Wendall} {lastName Griffin}]
              ...
 
@@ -474,5 +474,4 @@ their related reference and API documentation.
              :language: go
              :visible: false 
 
-             [{"firstName": "Emily" , "Description": "I love to play sports and
-             walk my beagle."} ... ]
+             [{"firstName": "Emily" , "Description": "I love to play sports and walk my beagle."} ... ]
--- a/source/quick-reference.txt
+++ b/source/quick-reference.txt
@@ -245,7 +245,7 @@ their related reference and API documentation.
              :language: go
              :visible: false
 
-             [{type ABC}, {xValue 5} ... ]
+             [{point ABC}, {xValue 5} ... ]
              [{point DCE},  ... ]
 
    * - | **Monitor Data Changes**
@@ -442,8 +442,8 @@ their related reference and API documentation.
              :language: go
              :visible: false 
 
-             [{type XYZ} {xValue 10}]
-             [{type DEF} {xValue 8}]
+             [{point XYZ} {xValue 10}]
+             [{point DEF} {xValue 8}]
              ...
 
    * - | **Create an Index**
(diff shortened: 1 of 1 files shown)


source/quick-reference.txt
```diff
@@ -28,13 +28,13 @@ their related reference and API documentation.
           .. input::
              :language: go
 
-             err = coll.FindOne(context.TODO(), bson.D{{"rating", 5}}).Decode(&result)
+             err = coll.FindOne(context.TODO(), bson.D{{"xValue", 5}}).Decode(&result)
```
rustagir, line 31: S: I think that this example is still a little difficult to follow. I noticed that elsewhere on this page you used sample documents that includes names -- maybe you could also use that here with an age filter or something!

Suggestion applies to all examples that use this sample data

source/quick-reference.txt
```diff
@@ -434,16 +434,16 @@ their related reference and API documentation.
                  context.TODO(), 
                  bson.D{},
                  options.Find().SetProjection(
-                     bson.D{{"vendor", 0}, {"_id",0}}
+                     bson.D{{"point", 0}, {"_id",0}}
```
rustagir, line 437: I: the projection sets `point` to 0 but the output still includes the field
S: omit the unprojected field from the output

rustagir (changes requested):
good work so far! i think some of the examples should be changed to be more intuitive. if you want to use abstract examples, try to make them as basic as possible and have the fields be `x`, `y`, `z`, or something like that, instead of matching points to fields. users could get confused why point `DEF` has an `xValue` for example.

source/quick-reference.txt
```diff
@@ -131,17 +131,16 @@ their related reference and API documentation.
 
              result, err := coll.UpdateMany(
                  context.TODO(), 
-                 bson.D{{"rating", bson.D{{"$lt", 10}}}}, 
-                 bson.D{{"$inc", bson.D{{"rating", 2}}}}
+                 bson.D{{"age", bson.D{{"$lt", 10}}}}, 
+                 bson.D{{"$set", bson.D{{"age", 1}}}}
```
rustagir, line 135: S: there's nothing wrong with this per say, but it might be confusing for readers to see an example where youre setting the age to 1 just because its less than 10. no need to change, but just suggesting

source/quick-reference.txt
```diff
@@ -434,16 +433,16 @@ their related reference and API documentation.
...
                  )
              )
 
           .. output::
              :language: go
              :visible: false 
 
-             [{type Masala} {rating 10}]
-             [{type English Breakfast} {rating 8}]
+             [{firstName Lester} ]
```
rustagir, line 444:
```suggestion
             [{firstName Lester}]
```

source/quick-reference.txt
```diff
@@ -469,10 +468,11 @@ their related reference and API documentation.
...
-             cursor, err := coll.Find(context.TODO(), bson.D{{"$text", bson.D{{"$search", "Breakfast"}}}})
+             cursor, err := coll.Find(context.TODO(), bson.D{{"$text", bson.D{{"$search", "beagle"}}}})
 
           .. output::
              :language: go
              :visible: false 
 
-             [{type English Breakfast} {rating 8} ... ]
+             [{"firstName": "Emily" , "Description": "I love to play sports and
+             walk my beagle."} ... ]
```
rustagir, line 478: S: make one line

source/quick-reference.txt
```diff
@@ -469,10 +468,11 @@ their related reference and API documentation.
...
              // only searches fields with text indexes
-             cursor, err := coll.Find(context.TODO(), bson.D{{"$text", bson.D{{"$search", "Breakfast"}}}})
+             cursor, err := coll.Find(context.TODO(), bson.D{{"$text", bson.D{{"$search", "beagle"}}}})
 
           .. output::
              :language: go
              :visible: false 
 
-             [{type English Breakfast} {rating 8} ... ]
+             [{"firstName": "Emily" , "Description": "I love to play sports and
```
rustagir, line 477:
```suggestion
             [{"firstName": "Emily" , "description": "I love to play sports and
```

rustagir (approved):
lgtm! good work responding to my comments.

PR #298, opened 2023-10-03
DOCSP-33282: Network compression

norareidy:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-33282>
Staging - <https://preview-mongodbnorareidy.gatsbyjs.io/golang/DOCSP-33282-network-compression/fundamentals/network-compression/>

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/network-compression.txt
+++ b/source/fundamentals/network-compression.txt
@@ -94,3 +94,20 @@ You do not need to add any dependencies to use the Zlib compression
 algorithm. Instead, import the built-in ``zlib`` package by adding ``import 
 "compress/zlib"`` to application files that specify a ``Client`` with
 Zlib compression.
+
+Additional Information
+----------------------
+
+For more information about the concepts in this guide, see the following
+documentation:
+
+- :ref:`golang-connection-uri`
+- :manual:`Connection String Compression Options </reference/connection-string/#compression-options>`
+  in the Server manual
+- `zlib <https://pkg.go.dev/compress/zlib>`__
+
+API Documentation
+~~~~~~~~~~~~~~~~~
+
+- `SetCompressors() <{+api+}/mongo/options#ClientOptions.SetCompressors>`__
+- `ClientOptions <{+api+}/mongo/options#ClientOptions>`__
--- a/source/fundamentals/network-compression.txt
+++ b/source/fundamentals/network-compression.txt
@@ -20,9 +20,9 @@ The {+driver-short+} supports the following compression algorithms:
 If you specify multiple compression algorithms, the driver selects the
 first one in the list supported by your MongoDB deployment.
 
-If you use the Snappy or Zstandard compression algorithm, you must add explicit
-dependencies. For more information, see the :ref:`compression-dependencies` section
-of this guide.
+You must add dependencies to use the Snappy or Zstandard compression algorithm.
+For more information, see the :ref:`compression-dependencies` section of this
+guide.
 
 .. _enable-compression:
 
@@ -33,7 +33,7 @@ You can enable compression for the connection to your MongoDB deployment
 by specifying the algorithms in one of two ways: 
 
 - Set the compression algorithm in your connection string.
-- Set the compression algorithm in a ``ClientOptions``  instance.
+- Set the compression algorithm in a ``ClientOptions`` instance.
 
 .. tabs::
 
@@ -90,10 +90,13 @@ following code:
 
    go get -u github.com/klauspost/compress
 
-You do not need to add any dependencies to use the Zlib compression
-algorithm. Instead, import the built-in ``zlib`` package by adding ``import 
-"compress/zlib"`` to application files that specify a ``Client`` with
-Zlib compression.
+To add the Zlib compression algorithm to your application, import the built-in
+``zlib`` package. You must add the following import statement to application files
+that instantiate a ``Client`` with Zlib compression:
+
+.. code-block:: go
+
+   import "compress/zlib"
 
 Additional Information
 ----------------------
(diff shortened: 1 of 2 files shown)


source/fundamentals/network-compression.txt
```diff
@@ -0,0 +1,97 @@
...
+   When using the Snappy or Zstandard compression algorithm, you must
+   :ref:`add explicit dependencies <compression-dependencies>`.
+
+.. _enable-compression:
+
+Specify Compression Algorithms
+------------------------------
+
+You can enable compression for the connection to your MongoDB instance
+by specifying the algorithms in one of two ways:
```
ccho-mongodb, line 34: Suggestion:
I think it could be good to use something similar to the [All List Items Are Sentences](https://www.mongodb.com/docs/meta/style-guide/style/lists/#all-list-items-are-sentences) style guide example to introduce the algorithms:
> You can do only one of the following:
>   Create multiple private endpoints in a single region.
>   Create one private endpoint in multiple regions.

E.g.
You can enable compression for the connection in one of the following ways:
- Add ...
- Specify ...

source/fundamentals/network-compression.txt
```diff
@@ -0,0 +1,97 @@
...
+.. _enable-compression:
+
+Specify Compression Algorithms
+------------------------------
+
+You can enable compression for the connection to your MongoDB instance
+by specifying the algorithms in one of two ways: 
+
+1. Adding the ``compressors`` parameter to your connection string. 
+2. Using the ``SetCompressors()`` method to specify a ``ClientOptions`` instance.
```
ccho-mongodb, line 37: Issue:

I think the style guide prefers unordered lists unless the items need to be sequential.

From the style guide on [Lists](https://www.mongodb.com/docs/meta/style-guide/style/lists/#list-items):
> Ordered lists or numbered lists. The list items must be performed or considered in a particular order.

Also, I think the second list item doesn't clearly communicate what the user needs to do.

Suggestion:

I think using unordered lists and the imperative mood could make these more compliant with the style guide and concise. Perhaps something like the following would be more concise and sufficiently descriptive:

```
- Set the compression algorithm in your connection string.
- Set the compression algorithm in a ``ClientOptions``  instance.
```

source/fundamentals/network-compression.txt
```diff
@@ -0,0 +1,97 @@
...
+
+.. tabs::
+
+   .. tab:: Connection String
+      :tabid: connection-string
+
+      To enable compression using the connection string, add the
```
ccho-mongodb, line 44: Issue:

Style guide on [using](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#std-term-using):
> by using with Use by using, or recast the sentence. Don't use using (alone) or with. By using is easier for worldwide audiences to understand.

Suggestion:

I think "by using" could replace "using".

```suggestion
      To enable compression by using the connection string, add the
```

Applies to all instances.

source/fundamentals/network-compression.txt
```diff
@@ -0,0 +1,97 @@
...
+      ``compressors`` parameter to your connection string. You can
```
ccho-mongodb, line 45: Suggestion:

I think since specifying compression requires both the compressors parameter and the compression algorithm, that information should be presented together.

```suggestion
      compression algorithm as the value of the ``compressors`` parameter to your connection string. You can
```

source/fundamentals/network-compression.txt
```diff
@@ -0,0 +1,97 @@
...
+            client, _ := mongo.Connect(context.TODO(), opts)
+
+   .. tab:: SetCompressors()
+      :tabid: mongoclientoptions
+
+      To enable compression using `SetCompressors()
+      <{+api+}/mongo/options#ClientOptions.SetCompressors>`__, pass
+      the compression algorithm you want to use to the ``SetCompressors()``
+      method. You can specify one or more compression algorithms separated
+      by commas:
```
ccho-mongodb, line 61: Issue:
I think this is missing some important information. In the code example, it shows that they must be elements of a string array, which happens to be defined inline by using commas, but could've been added in many other ways. 

Suggestion:
I think it would be more accurate to mention the type that SetCompressors() accepts (appears to be a string array) than to mention that they must be separated by commas.

source/fundamentals/network-compression.txt
```diff
@@ -0,0 +1,97 @@
...
+following code: 
+
+.. code-block:: bash
+
+   go get -u github.com/klauspost/compress
+
+.. note::
+
+   To use the Zlib compression algorithm, import the built-in Go package
+   ``compress/zlib`` in your application files.
```
ccho-mongodb, line 97: Suggestion:
I think including Zlib instructions in a callout makes the information more difficult to find than if it was in normal text, similar to the other libraries.

I also think it could help to identify which application files in which to add the compress/zlib import.

ccho-mongodb (changes requested):
Looks pretty good. The item about the compress/zlib import wasn't that clear to me so I left a suggestion. I would like to take another look after you make changes, mostly for that part.

ccho-mongodb (approved):
LGTM! I identified a couple minor items that would be good to to address before merging.

source/fundamentals/network-compression.txt
```diff
@@ -18,47 +18,46 @@ The {+driver-short+} supports the following compression algorithms:
...
 
 If you specify multiple compression algorithms, the driver selects the
-first one in the list supported by your MongoDB instance.
+first one in the list supported by your MongoDB deployment.
 
-.. note::
-
-   When using the Snappy or Zstandard compression algorithm, you must
-   :ref:`add explicit dependencies <compression-dependencies>`.
+If you use the Snappy or Zstandard compression algorithm, you must add explicit
```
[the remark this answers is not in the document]
norareidy: Reworded this part

PR #342, opened 2024-01-25
DOCSP-35814: renames connid64 type, removes old field in v2

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-35814>
Staging - https://preview-mongodbrustagir.gatsbyjs.io/golang/DOCSP-35814-serverconnID-field-rm/whats-new/#std-label-version-2.0

## Self-Review Checklist

- [x] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?

### DIFF
--- a/source/fundamentals/monitoring/command-monitoring.txt
+++ b/source/fundamentals/monitoring/command-monitoring.txt
@@ -86,7 +86,6 @@ CommandStartedEvent
        "RequestID": ...,
        "ConnectionID": "...",
        "ServerConnectionID": ...,
-       "ServerConnectionID64": ...,
        "ServiceID": "..."
    }
 
@@ -104,7 +103,6 @@ CommandSucceededEvent
        "RequestID": 13,
        "ConnectionID": "...",
        "ServerConnectionID": ...,
-       "ServerConnectionID64": ...,
        "ServiceID": null,
        "Reply": "..."
    }
@@ -123,7 +121,6 @@ CommandFailedEvent
        "RequestID": 13,
        "ConnectionID": "...",
        "ServerConnectionID": ...,
-       "ServerConnectionID64": ...,
        "ServiceID": null,
        "Failure": "..."
    }
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -57,10 +57,17 @@ What's New in 2.0
 
 The 2.0 {+driver-short+} release includes the following improvements and fixes:
 
-- Consolidation of the ``ServerConnectionID64`` and ``ServerConnectionID``
-  fields of the ``CommandStartedEvent`` and ``CommandFinishedEvent``
-  structs into the ``ServerConnectionID`` field. This field in all
-  ``PoolEvent`` documents has a value of type ``int64``.
+- Updates to monitoring event documents:
+
+  - Consolidation of the ``ServerConnectionID64`` and ``ServerConnectionID``
+    fields of the ``CommandStartedEvent`` and ``CommandFinishedEvent``
+    structs into the ``ServerConnectionID`` field, which takes a value of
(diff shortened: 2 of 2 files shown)


source/fundamentals/monitoring/command-monitoring.txt
mongoKart, line 1: This ticket is confusing. The first part seems to say that `ServerConnectionID` should be removed, but the Doc Changes Summary says the `ID64` should be removed. The ticket also mentions only `CommandStarted` and `CommandFinished`, but clearly those fields are in more places than that.

Just wanted to check that what you have here is right!

source/whats-new.txt
```diff
@@ -55,6 +55,19 @@ What's New in 2.0
    :ref:`Connection Example Code <go-connection-example-code>` in the
    Connection Guide.
 
+The 2.0 {+driver-short+} release includes the following improvements and fixes:
+
+- Updates to monitoring event documents:
+
+  - The ``CommandStartedEvent`` and ``CommandFinishedEvent`` structs
```
mongoKart, line 62: Q: Should this be `CommandFailedEvent` to match the change made above? Should it include `CommandSucceededEvent`?

source/whats-new.txt
```diff
@@ -55,6 +55,19 @@ What's New in 2.0
...
+    have a single ``ServerConnectionID`` field of type ``int64`` to
+    capture the connection ID.
+    
+  - Changes the type of the ``ConnectionID`` field in ``PoolEvent``
+    structs to ``int64`` from ``uint64``.
```
mongoKart, line 67: I: [Start all items with fragments or all items with sentences](https://www.mongodb.com/docs/meta/style-guide/style/lists/#list-items)

```suggestion
  - The ``ConnectionID`` field in ``PoolEvent``
    structs now use the ``int64`` data type instead of ``uint64``.
```

mongoKart (changes requested):
one issue and a couple questions!

source/fundamentals/monitoring/command-monitoring.txt
rustagir: My understanding is that there used to be two fields: `ServerConnectionID` and `ServerConnectionID64`. `ServerConnectionID` was deprecated, but in this release that functionality is removed but `ServerConnectionID64` is RENAMED to `ServerConnectionID` to serve the same purpose essentially. So even though `ServerConnectionID` is removed, by appearance it looks like  `ServerConnectionID64` is removed

 The other part is that the two of the three command events (succeeded, failed) use the same template struct (CommandFinished), with one extra field each. If it makes it clear, I can replace CommandFinished with the two "finished" events instead, even though the code change is really to CommandFinished, not to the two event structs.

source/whats-new.txt
rustagir: see my comment

source/whats-new.txt
```diff
@@ -55,6 +55,19 @@ What's New in 2.0
...
+The 2.0 {+driver-short+} release includes the following improvements and fixes:
+
+- Updates to monitoring event documents:
+
+  - The ``CommandStartedEvent`` and ``CommandFinishedEvent`` structs
+    have a single ``ServerConnectionID`` field of type ``int64`` to
+    capture the connection ID.
+    
+  - The ``ConnectionID`` field in the ``PoolEvent``
+    struct take a value of type ``int64`` instead of ``uint64``.
```
mongoKart, line 67: I: Should be "fields" in the first line or "takes" in the second

mongoKart (approved):
lgtm after one fix

matthewdale (approved):
Looks good! 👍

PR #474, opened 2025-03-11
DOCSP-46818: Client bulk write

norareidy:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-46818
Staging - https://deploy-preview-474--docs-golang.netlify.app/fundamentals/crud/write-operations/bulk/

## Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?
- [ ] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?


source/fundamentals/crud/write-operations/bulk.txt
```diff
@@ -15,37 +15,55 @@ Bulk Operations
...
-making a call for each operation to the database, bulk operations
-perform multiple operations with one call to the database.
+The ``Collection`` and ``Client`` classes both provide a ``BulkWrite()``
+method. When calling ``BulkWrite()`` on a ``Collection`` instance, you can
+perform multiple write operations on a single collection. When calling
+``BulkWrite()`` on a ``Client`` instance, you can perform bulk writes across
+multiple namespaces. In MongoDB, a namespace consists of the database name and the collection
+name in the format ``<database>.<collection>``. 
+
+.. important::
```
rustagir, line 30: S: add a title
"Client Bulk Write Compatibility"

rustagir (changes requested):
As I was reviewing I noticed that you have organized this page into functions instead of APIs. Personally, I found it more difficult to move through sections, then find my specific API of interest as opposed to organizing them into two separate sections as we do [here](https://www.mongodb.com/docs/languages/kotlin/kotlin-sync-driver/current/write/bulk-write/) . Happy to talk more about how to best organize the page but I think it is confusing to stack the collection/client info in each section

source/fundamentals/crud/write-operations/bulk.txt
```diff
@@ -15,54 +15,73 @@ Bulk Operations
...
 
-Bulk operations perform a large number of write operations. Instead of
-making a call for each operation to the database, bulk operations
-perform multiple operations with one call to the database.
+The ``Collection`` and ``Client`` classes both provide a ``BulkWrite()``
+method. When calling ``BulkWrite()`` on a ``Collection`` instance, you can
+perform multiple write operations on a single collection. When calling
+``BulkWrite()`` on a ``Client`` instance, you can perform bulk writes across
+multiple namespaces. In MongoDB, a namespace consists of the database name and the collection
+name in the format ``<database>.<collection>``.
```
rustagir, line 28: S: in the bulk write method the namespaces are not given in that format so I recommend removing to avoid confusion
```suggestion
method. You can use the ``Collection.BulkWrite()`` method to
perform multiple write operations on a single collection. When calling
the ``Collection.BulkWrite()`` method, you can perform bulk writes across
multiple namespaces. In MongoDB, a namespace consists of a database name and a collection name. 
```

source/fundamentals/crud/write-operations/bulk.txt
```diff
@@ -15,54 +15,73 @@ Bulk Operations
...
+multiple namespaces. In MongoDB, a namespace consists of the database name and the collection
+name in the format ``<database>.<collection>``. 
+
+.. important:: Client Bulk Write Requirements
+
+   To perform bulk operations on a ``Client`` instance,
+   ensure that your application meets the following
+   requirements:
+
+   - Uses {+driver-short+} v2.1 or later
```
rustagir, line 36: S: remove this first pooint and combine into a sentence. This change will only be on the v2.1+ branch anyways so no need to explicitly include

source/fundamentals/crud/write-operations/bulk.txt
```diff
@@ -15,54 +15,73 @@ Bulk Operations
...
-~~~~~~~~~~~~~~~
+To perform a bulk operation on a single namespace, call the ``BulkWrite()``
+method on a collection and pass an array of :ref:`WriteModel <golang-write-model-collection>`
+documents as a parameter.
 
-The ``BulkWrite()`` method optionally takes a ``BulkWriteOptions``
-type, which represents options you can use to modify its behavior. If
-you don't specify a ``BulkWriteOptions``, the driver uses the default
-values for each option.
+Modify Collection-Level Behavior
```
rustagir, line 78: S suggest moving the options and return value sections to the end of this section after the define write models section (applies to collection and client)

source/fundamentals/crud/write-operations/bulk.txt
```diff
@@ -226,33 +248,36 @@ their behavior with the following methods:
    * - ``SetHint()`` 
      - | The index to use to scan for documents.
 
+   * - ``SetSort()`` 
+     - | The criteria to use when ordering matching documents.
+         This method is not available for the ``UpdateMnyModel``
```
rustagir, line 253:
```suggestion
         This method is only available for the ``UpdateOneModel``
```

source/fundamentals/crud/write-operations/bulk.txt
```diff
@@ -366,6 +370,346 @@ the bulk operation:
...
+     - A boolean value that indicates whether the result contains detailed results.
+
+.. _golang-write-model-client:
+
+Define Client Bulk Write Models
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To specify the write operations for your bulk operation on multiple
+namespaces, create a ``ClientWriteModel`` for each insert,
+replace, update, or delete.
```
rustagir, line 476: S: add a sentence describing how the first two parameters for each model supply the target namespace

source/fundamentals/crud/write-operations/bulk.txt
```diff
@@ -366,6 +370,346 @@ the bulk operation:
...
+
+- Update operation on the ``poems`` collection to update a 
+  document in which the ``author`` value is ``"Ada Limon"``
+
+.. literalinclude:: /includes/fundamentals/code-snippets/CRUD/bulkOps.go
+   :language: go
+   :dedent:
+   :start-after: begin bulk update model client
+   :end-before: end bulk update model client
+
```
rustagir, line 616: S: add a note that describes updatemany model as an option to update all matching documents

source/fundamentals/crud/write-operations/bulk.txt
```diff
@@ -390,13 +734,21 @@ API Documentation
 To learn more about any of the methods or types discussed in this
 guide, see the following API Documentation:
 
-- `BulkWrite() <{+api+}/mongo#Collection.BulkWrite>`__
```
rustagir, line 393: S: consider sorting this into two lists as done on the kotlin sync page

rustagir (changes requested):
A few things but great work fixing up this page!

source/whats-new.txt
```diff
@@ -26,7 +26,11 @@ What's New
 
 Learn what's new in:
 
+<<<<<<< HEAD
+* :ref:`Version 2.1 <golang-version-2.0>`
+=======
 * :ref:`Version 2.1 <golang-version-2.1>`
+>>>>>>> upstream/master
```
rustagir, line 33: S: I believe there's a merge error still in the PR

source/fundamentals/crud/write-operations/bulk.txt
```diff
@@ -73,19 +262,20 @@ following methods:
...
-~~~~~~~~~~~~~
+Collection-Level Return Value
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The ``BulkWrite()`` method returns a ``BulkWriteResult`` type, which
-contains information about the bulk operation if it's successful. The
-``BulkWriteResult`` type contains the following properties:
+includes information about the bulk operation.
+
+The ``BulkWriteResult`` type contains the following properties:
```
qingyang-hu, line 278: Do we need to add the `Acknowledged` property?

norareidy: added!

source/fundamentals/crud/write-operations/bulk.txt
```diff
@@ -73,19 +262,31 @@ following methods:
...
 
+   * - ``SetComment()``
+     - | Specifies a comment to attach to the operation. 
+       | Default: ``nil``
+
+   * - ``SetLet()``
+     - | Specifies a document with a list of values to improve operation readability. Values
+         must be constant or closed expressions that don't reference document fields. For more
+         information, see the :manual:`let statement
+         </reference/command/delete/#std-label-delete-let-syntax>` in the {+mdb-server+} manual.
```
qingyang-hu, line 276: Do we need to mention the </reference/command/update/#std-label-update-let-syntax> as well?
Also, do we need to somehow align this section with the `SetLet` in the "Client-Level Behavior" underneath?

qingyang-hu: 👍 IMO, a link to a generic reference is not necessary for the Golang-specific doc, neither the delete nor the update. Moreover, listing the delete syntax alone causes a feeling like it's for the delete operation only. However, I'm OK with either keeping the "delete-let-syntax" link or removing it. @jtazin, what are your thoughts?

qingyang-hu (approved):

source/fundamentals/crud/write-operations/bulk.txt
jtazin: Yeah, I agree that it could be confusing to link to one and not the other. It sounds like it would be best to either have both or none.
I do like having links for more information, especially since "let" doesn't seem obvious to me.
The [Server bulk write documentation](https://www.mongodb.com/docs/manual/reference/command/bulkWrite/) includes both. To bad there isn't a way to link directly to the "let" on this page. Perhaps we can copy what we do?
Either way, I don't feel too strongly about this one.
@norareidy and @qingyang-hu let me know if you have additional thoughts.

